PR #2, opened 2026-04-12
Fix cargo target dir guidance

mevanlc:
## Summary
- stop telling local host developers to export `CARGO_TARGET_DIR=/target`
- document that local development can use Cargo's default repo-local `target/`
- clarify that `/app/target` is the Docker/dev-container path
- explicitly ignore `target/` in the workspace root

## Notes
- no code behavior changes
- docs/config guidance only

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-target
+target/
 node_modules
 .env
 .claude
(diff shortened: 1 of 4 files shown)


mpiorowski:
Thank you! :) First contribution <3

PR #3, opened 2026-04-12
Compact music pairing session tokens

mevanlc:
## Summary
- encode music pairing session tokens as unpadded urlsafe base64 over UUID v7 bytes
- keep the token flow opaque-string compatible while shortening connect URLs
- update context docs for the compact token handshake format

## Verification
- cargo fmt --all
- CARGO_TARGET_DIR=target CARGO_HOME=$HOME/.cargo cargo check -p late-ssh -p late-cli


mpiorowski (approved):
goat

PR #5, opened 2026-04-13
Add /ignore and /unignore chat commands

mevanlc:
## Summary

- Add `/ignore [username]` and `/unignore [username]` chat commands for per-user muting
- Ignore list is persisted in the user's JSON `settings` column and survives reconnects
- Messages from ignored users are filtered client-side (both on snapshot load and real-time push); mentions/notifications are not affected

Run `/ignore` or `/unignore` without an argument to see your ignore list.

TBD:  when user picks a new nickname, update all users ignore list to track that username? 

## Design

**Persistence** (`late-core/src/models/user.rs`): ignore list stored as `ignored_usernames` in the user settings JSON. CRUD via `add_ignored_username` / `remove_ignored_username` / `ignored_usernames`. An `IgnoreListMutation` enum gives callers enough detail for user-facing feedback (Added, AlreadyPresent, Removed, Missing).

**Service** (`late-ssh/src/app/chat/svc.rs`): `ignore_user_task` / `unignore_user_task` validate the target exists and isn't self, delegate to the model, then broadcast `IgnoreListUpdated` or `IgnoreFailed` events. The snapshot includes `ignored_usernames` so the list loads on connect.

**Client state** (`late-ssh/src/app/chat/state.rs`): holds `ignored_usernames: HashSet<String>`. Filters messages via `message_is_ignored` at two points: `push_message` (real-time) and `merge_rooms` (history load). Also extracts `clear_composer_after_submit()` to deduplicate the pattern that was repeated 8+ times.

## Review notes

- Usernames are normalized (lowercased, trimmed, leading `@` stripped) and deduplicated on read via `BTreeSet`
- Self-ignore is prevented server-side
- Settings read-modify-write preserves other fields (e.g. `"theme"`) — tested explicitly
- Mentions/notifications flow through a separate subsystem and are unaffected by ignores
- DM blocking is client-side filtering: messages from ignored users in DM rooms are hidden, but the DM room itself remains visible (existing DMs are not deleted)
- On ignore list change, a full `request_list()` re-fetch ensures all visible message lists are re-filtered

## Test plan

- [x] Unit tests for `/ignore` and `/unignore` command parsing (with/without `@`, bare command lists, non-matching input)
- [x] Integration tests for model-layer normalization, dedup, add/remove/already-present/missing mutations, settings preservation
- [x] Service event tests for ignore/unignore task emit, self-ignore rejection, duplicate rejection, missing user errors
- [x] Snapshot test verifying persisted ignores appear in the initial snapshot
- [x] End-to-end app input flow test: ignore command hides messages and persists


mpiorowski:
amazing pr, used it as baase for the overlay system! Kept most of the logic, was easier to work on my branch :) !!! YOU ARE GOAT

PR #8, opened 2026-04-13
Swap h/l room navigation for LTR-friendly ordering

mevanlc:
## Summary
- Swap the direction of `h` and `l` for room switching: `l` now moves to the next room (down the list) and `h` moves to the previous room (up the list)
- Update all hint texts from `h/l` to `l/h` to reflect the new ordering

## Motivation
In vim (the inspiration for hjkl movement), l = right and h = left.  For LTR language readers (the majority of late.sh users), "right" intuitively means "next" and "left" means "previous". When working with a horizontal room list whose selection starts at the top, `l` (right) makes more sense to go "down" to the next room ordinally, and `h` (left) to go back to the previous one.

## Test plan
- [ ] Verify `l`/`L` moves to the next room in the list
- [ ] Verify `h`/`H` moves to the previous room in the list
- [ ] Confirm hint text reads `l/h` in the Rooms block title, `/help` output, and test assertions
- [ ] Check room switching still works from notifications and news views

### DIFF
--- a/late-ssh/src/app/chat/input.rs
+++ b/late-ssh/src/app/chat/input.rs
@@ -79,11 +79,11 @@ pub fn handle_byte(app: &mut App, byte: u8) -> bool {
     if app.chat.notifications_selected {
         match byte {
             b'h' | b'H' => {
-                switch_room(app, 1);
+                switch_room(app, -1);
                 return true;
             }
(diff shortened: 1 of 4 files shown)


mpiorowski:
Ive done it already! :D Pushed to main ;p

But thank you!

PR #9, opened 2026-04-13
Use stateful VT parsing for SSH input and improve chat navigation

mpiorowski:
## Summary
- Replaces custom input decoding with a robust, stateful VT parser (`vte`) to correctly handle split escape sequences and bracketed pastes across partial SSH network reads.
- Improves chat message management and navigation UX by keeping message selection active after deletion and tuning paging scroll distances.

## What changed
- **Stateful VT Parsing:** SSH input is now routed through a persistent `vte::Parser`. This prevents issues where escape sequences or `[200~` paste markers could be fragmented across TCP packets and mistakenly parsed as literal keystrokes or newlines.
- **Continuous Message Deletion:** Pressing `d` to delete a message now selects the adjacent message (preferring the next older message) instead of clearing the selection. This allows users to rapidly delete a run of their own messages with repeated presses.
- **Intuitive Room Switching:** Swapped the directions for `h` (left) and `l` (right) room switching to correctly match standard vim directional expectations.
- **Tuned Paging:** Ctrl-U and Ctrl-D now scroll by a calculated distance that accounts for average message wrap height (assuming ~3 rows per message) so jumps feel more accurately like half a visible page.
- **Internal Cleanup:** Removed an unused database dependency from `BlackjackService::new`.

## Behavior notes
- A standalone `Esc` key press is now resolved using a short 40ms tick delay. This ensures we don't prematurely exit composing modes if the `Esc` byte was actually the beginning of a control sequence (like an arrow key) that was split across network reads.

## Feature flags
None

## Screenshots / Video
N/A - This PR does not change the visual UI, only input handling and interaction mechanics.

## Testing
- Added and updated automated unit tests to verify the stateful behavior of `VtInputParser`, split sequence reconstruction, and the custom `split_alt_enter` parsing.
- Automated tests added for the new `adjacent_message_id` logic to ensure boundary conditions behave correctly (e.g., deleting the oldest or newest message).

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -547,6 +547,7 @@ Known gaps/risks:
 - Time remaining is approximate (up to 5s polling delay on track change)
 - No external metrics or alerting system
 - **Single-replica assumption:** Several structures are purely in-memory and not shared across processes (see multi-replica notes below)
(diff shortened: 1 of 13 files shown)


mevanlc:
Ooh good. I was having problems with bracketed paste & unbracketed paste in my unit tests. Will check this out.

PR #10, opened 2026-04-13
Add users list tab mockups

mevanlc:
## Summary
- Add drawio mockups for the users list tab (tab1 and tab1-touchup)

## Test plan
- [ ] Open `.drawio` files in draw.io to verify they render correctly


mpiorowski:
Thank you, but initially we are going with the `/active` command :) more space in the TUI!

PR #12, opened 2026-04-13
Made CLI Work on Windows

SimonSeider:
Adds Windows support to the late-cli binary, enabling it to compile and build on Windows systems.
### Changes
- Add conditional compilation (#[cfg(unix)] vs #[cfg(not(unix))]) for platform-specific code
- Windows SSH spawns directly without PTY forwarding
- Update Cargo.toml with valid edition = "2021" 
- Add platform-specific SshProcess and PtyResizeHandle structs
- Fix ssh_dir() to use USERPROFILE on Windows
- Add Windows terminal size detection fallback
### Notes
- On Windows, SSH runs directly in terminal without stdin/stdout streaming
- Token capture and full WS pairing features work on Unix only in this version (will be implemented later)
- Build: cargo build -p late-cli --target x86_64-pc-windows-msvc
### Testing
Build verified with cargo check -p late-cli


mpiorowski (changes requested):
This PR is not full Windows support yet. Right now it mostly adds cfg guards so the CLI can move toward compiling on Windows, and it starts an ssh process there, but the actual Windows session flow is still missing.

Right now this is basically just launching ssh, like running \ssh late.sh manually. What’s still missing for real Windows support is the late-cli integration around it: reading ssh output, extracting the session token, forwarding input properly, and starting the pairing/audio websocket flow.

@ExilProductions wanna try to improve this one? :)

SimonSeider:
yeah i know i got it working after this commit, ima commit the changes soon

PR #14, opened 2026-04-13
Allow Escape to skip the intro splash animation

mevanlc:
## Summary
- Pressing Escape while the coffee-cup intro animation is playing now skips the remainder and transitions directly to the welcome overlay
- Same end-state as letting the animation run to completion (splash_ticks > 90)

## Test plan
- [ ] SSH in and let the splash animation play normally — verify it still runs to completion
- [ ] SSH in and press Escape mid-animation — verify it skips to the welcome overlay immediately
- [ ] Press a non-Escape key during splash — verify it is still ignored (no skip)


mpiorowski:
toooo much conflicts, will extract the func into a separate pr :D

mevanlc:
> toooo much conflicts, will extract the func into a separate pr :D

@mpiorowski Please do this one asap, because I do manual testing and waiting for the splash screen every time I want to test a code change is annoying. <.<

Sry about the conflicts, this branch got messed up.

Edit: I see you already took the code into your main, thx.

PR #17, opened 2026-04-13
Add per-user chat input history with up/down arrow navigation

mevanlc:
## Summary
- Adds in-memory per-user chat input history (last 20 messages) navigable with Up/Down arrows while composing, for both SSH TUI and web chat
- Up steps through older entries, Down steps toward newer; navigation stops at boundaries without looping
- Pressing Down on unsent text records it to history and clears the input; Down on empty input does nothing
- Draft text is saved when entering history navigation and restored when exiting, so accidental Up/Down doesn't lose in-progress work
- Editing a recalled history entry and sending creates a new entry; the original stays unchanged
- History stores `trim_end()` to match exactly what `submit_composer()` sends

## Implementation
- Extracts `HistoryNav` struct in `late-ssh/src/app/chat/state.rs` — self-contained state machine with `up()`, `down()`, `on_submit()` methods that take/return plain strings with no service dependencies
- `ChatState` delegates via `history_up()`, `history_down()`, and `clear_composer_after_submit()`
- Web chat mirrors the same logic in the Alpine.js `chat()` component with `messageHistory`, `historyIndex`, `historyDraft`
- Up/Down arrows in compose mode now drive history instead of multi-line cursor movement (`composer_cursor_up`/`composer_cursor_down`)

## Test plan
- [x] 13 unit tests for `HistoryNav` covering all edge cases (empty history, boundary stops, draft save/restore, cap at 20, whitespace handling, trim consistency)
- [ ] Manual: SSH — compose mode, send a few messages, Up/Down through history, verify draft preservation
- [ ] Manual: Web chat — same flow in browser
- [ ] Manual: verify sending an edited history entry creates a new entry without modifying the original


mpiorowski:
Thanks for this, and the bug fix + tests on top.

After consideration, even though I was the one who added the suggestion :D I'm gonna pass, sorry. Two things:

1. Arrow-key history is a shell habit. In chat people just retype or scroll up. Don't think it gets used enough to justify carrying.
2. Bigger problem, our composer is multi-line, and Up/Down already move the cursor between lines. This PR takes those keys away for history, which breaks something people actually use.

PR #19, opened 2026-04-15
Add chat message edit mode

mevanlc:
## Summary

Add author-only chat message editing in the TUI chat views.

This change lets a user select one of their own messages, press `e`, edit the message in the existing composer, and save it back to the original row instead of creating a new message.

## What Changed

- added edit mode for selected own chat messages in the main chat screen and dashboard chat card
- made Enter submit edits back to the selected message instead of sending a new message
- made `Esc` cancel edit mode on first press while keeping the composer focused, then exit compose on the second press
- rendered edited messages with an inline `· edited` marker in the message header
- added service events and state reconciliation for successful and failed message edits
- added service and input-flow coverage for edit mode behavior

## Why

The chat composer already supported sending new messages and replying to selected messages, but there was no path to correct or revise a message after sending it. That forced follow-up correction messages and made the selected-message keyboard workflow inconsistent.

## Impact

Users can now revise their own messages quickly from the keyboard without leaving the current chat flow. The feature stays author-only and uses the existing `updated` timestamp to signal edits in the UI.

## Validation

- `cargo fmt --all`
- `cargo clippy -p late-ssh --all-targets -- -D warnings`
- `cargo test -p late-ssh --lib`
- `cargo test -p late-ssh --test chat`
- `cargo test -p late-ssh --test app_input_flow`


mpiorowski:
Added in my char rewrite! We need better communication :D

PR #23, opened 2026-04-15
Make CLI/music instructions and vote panel responsive to narrowing

mevanlc:
## Summary

Made the voting box and music connection box responsive by having narrowing render states. This allows the Dashboard's chat to be usable on much narrower terminals than before (such as Termux, or older computer monitors, or for people that need larger fonts to read well.) The code is unfortunately complex for such a "simple" thing but it can always be backed if/when/as-soon-as it becomes a burden, so I'm in favor of trying it out.

## Detail
- Lower dashboard min width to 24 and stack vote card above chat when too narrow to split
- Add compact stream card layout showing just the Enter/P shortcuts when space is tight
- Add compact vote option rendering with accented initial letters and shortened card title
- Guard top-level draw and resize against zero-sized areas, render "Too small" below 4x4


mevanlc:
Handled by a different PR.

PR #25, opened 2026-04-16
Clean up expired per-IP entries in `IpRateLimiter`

andschneider:
Hey, awesome project! I've been working on my own `ratatui` + `russh` app recently, so I was excited to dig into this.

While reading through the rate limiter, I noticed that expired per-IP entries can stick around indefinitely in the `HashMap`, which means scanner traffic can accumulate in the map until restart. This PR adds a periodic cleanup to remove those expired entries.

## Problem

`IpRateLimiter::allow()` prunes expired timestamps from the current IP's `VecDeque`, but it does not remove the IP key from the outer `HashMap` when that deque becomes empty.

That means IPs that hit the limiter once and never call `allow()` again can leave behind empty map entries indefinitely. For an internet-facing service, scanner traffic can make that map grow monotonically until restart.

## Impact

- Rate-limiting behavior is unchanged. Empty entries do not affect correctness.
- Memory usage can grow without bound as new source IPs accumulate.
- Both `ssh_attempt_limiter` and `ws_pair_limiter` are affected.

## Fix

Add a periodic background cleanup task that sweeps each limiter's `attempts_by_ip` map, removes expired timestamps, and drops IP entries whose queues become empty.

This keeps the `O(n)` sweep off the request path. Expired entries may remain until the next cleanup tick, but map growth is bounded instead of unbounded.

## Verification

- Added a unit test that confirms cleanup removes an IP entry after its timestamps expire.
- Added a unit test that confirms cleanup preserves IPs with active timestamps.

### DIFF
--- a/late-core/src/rate_limit.rs
+++ b/late-core/src/rate_limit.rs
@@ -30,6 +30,24 @@ impl IpRateLimiter {
         self.window.as_secs()
     }
 
+    pub fn entry_count(&self) -> usize {
+        self.attempts_by_ip.lock_recover().len()
+    }
+
+    pub fn cleanup(&self) {
+        let now = Instant::now();
+        let mut attempts_by_ip = self.attempts_by_ip.lock_recover();
+        attempts_by_ip.retain(|_, attempts| {
+            while let Some(first) = attempts.front() {
+                if now.duration_since(*first) <= self.window {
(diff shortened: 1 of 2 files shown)


mpiorowski (approved):
Great catch, thank you!

PR #38, opened 2026-04-16
Adds Black Background Setting to provide contrast for users with alternate Terminal themes

Jshmill:
First of all - amazing project! I love it! My one problem with it is that I'm running a slightly transparent catppuccin themed terminal, and so whenever I open late.sh, the text (especially the dull brown) is extremely difficult to see. I've thought about adding themes but that seems like a big project and I'm still new to rust, so I figured having this setting for users to turn on would be the best place for me to start.

## Changes
- Added a toggle setting to allow users to toggle on and off a black background to improve readability

## Before:
Screenshot 2026-04-16 at 9 30 22 AM

## After:
Screenshot 2026-04-16 at 9 30 39 AM

Again - I'm new to rust, so a lot of this development was aided by AI, and while I studied and understand the code that I added, I think a little extra scrutiny would be good.

Overall, I think adding theming would be a perfect solution to let users select other themes (catppuccin, maybe their personal terminal theme, etc.), but I think that this could be a perfect alternative for now, and I do love the cozy theme.

Again, amazing job on the project!


mpiorowski (changes requested):
This look great, but unfortunately Ive just merged a Theme system.

Would you like to take another look, even with ai, to add it as part of the new theme setting? :)

mevanlc:
> This look great, but unfortunately Ive just merged a Theme system.
> 
> Would you like to take another look, even with ai, to add it as part of the new theme setting? :)

I suggest making BG color a separate setting from theme.  If not a color picker, then at least an on/off switch like "Use theme BG" [x]

Forcing BG color on all users makes some unhappy.

Jshmill:
@mpiorowski @mevanlc I like that, and yeah I can fix it to work with the themes and make them feel more immersive

mevanlc:
> I like that, and yeah I can fix it to work with the themes and make them feel more immersive

Maybe pick a light BG color that goes with each theme and also a dark BG color that goes with each theme. 

If theme simply cannot work with a true light or true dark background (due to contrast issues) then pick a slightly lighter dark one for Light, or a slightly darker light one for Light, depending on which doesn't work well.

Then user can pick: 
```
Theme Background:  ◀ <Off|Dark|Light> ▶
 ```
 
 @Jshmill thanks for contributing btw

Jshmill:
I fixed it to merge with themes, and now each theme has its own background color that can be toggled on and off. As of right now, the Light and Dark background might be difficult, because, for example, I think that the Late theme would not cater to a lighter background.

mevanlc:
> I fixed it to merge with themes, and now each theme has its own background color that can be toggled on and off. As of right now, the Light and Dark background might be difficult, because, for example, I think that the Late theme would not cater to a lighter background. 

You're right. Let's start with On/Off switch then?

mpiorowski:
@Jshmill TYHANK YOU, whats your username, write in the dev channel so i can mention you

PR #39, opened 2026-04-16
Narrow-terminal (mobile) dashboard and composer layout

mevanlc:
- adapt dashboard screen 1 by progressively removing vote and stream panels at tighter widths
- make composer title collapse stages explicit and responsive to terminal width


mpiorowski (changes requested):
Composer breakpoints should key off pane width, not terminal width

ComposerTitleStage::for_terminal_width receives self.size.0 (full terminal width), but the composer lives inside a pane that's terminal_width - 31 on the dashboard (outer border 2 + sidebar 24 + chat block border 2 + composer padding 3). The thresholds 82/74/57/50 look tuned for terminal width but actually encode this -31 offset implicitly — which is fragile and undocumented.

Two concrete problems:

1. Latent bug on the Chat screen. draw_chat splits content_area further into a room list + message pane, so the composer pane is narrower than terminal_width - 31. Thresholds over-estimate
  available width there and hints will overflow before they collapse.
2. Layout-coupled constants. Any change to sidebar width, border, or composer padding silently desyncs the thresholds. Nothing in ComposerTitleStage signals that dependency.

Suggested fix: rename terminal_width → composer_pane_width on ComposerBlockView, have draw_composer_block pass area.width (post-border), and rebase thresholds on pane width (~53/45/26/19). This matches the pattern already used for DASHBOARD_HIDE_VOTE_AT_WIDTH, which is correctly content-width-based.

Tests would also become meaningful end-to-end: passing 51 into composer_title would then correspond to an actual 51-col pane in production.

mevanlc:
FYI this is the direction I'm heading:

okay reviewer is making a reasonable request -- we really should be collapsing the input box's (IB's) title based on the input box's width and what it can display in the region available for its "titled border" -- and not the overall terminal width. Using the TW makes the layout brittle because if someone wants to rearrange components they have recalculate the IB's relationship to the new TUI structure. I'm thinking we should make a generic 'TitledBorderContainer' or something (I really don't know what to call it or how ratatui wants us to do things, but the idea is to componentize and delegate responsibility to more-local state rather than more-global state). TitledBorderContainer can be initialized with a list of strings that represent the collapse tiers of the title. It then selects the longest title which would not truncate.

mevanlc:
> FYI this is the direction I'm heading:
> 
> okay reviewer is making a reasonable request -- we really should be collapsing the input box's (IB's) title based on the input box's width and what it can display in the region available for its "titled border" -- and not the overall terminal width. Using the TW makes the layout brittle because if someone wants to rearrange components they have recalculate the IB's relationship to the new TUI structure. I'm thinking we should make a generic 'TitledBorderContainer' or something (I really don't know what to call it or how ratatui wants us to do things, but the idea is to componentize and delegate responsibility to more-local state rather than more-global state). TitledBorderContainer can be initialized with a list of strings that represent the collapse tiers of the title. It then selects the longest title which would not truncate.

@mpiorowski 

Done and requesting re-review.  Substantially more code changed but this sets us up with a reusable pattern for collapsing titles strings (and such) responsively across the app.  Considering making it more reusable (not just a function pattern), but we'll see if it's YAGNI.

PR #42, opened 2026-04-16
Fix cannot type j or k into emoji picker search bar

mevanlc:
- Replace plain j/k navigation with Ctrl+J/Ctrl+K (0x0B/0x0A)
- Allows typing j and k normally in the search box


mpiorowski (changes requested):
Change the desc inside /help also :)

mevanlc:
> Change the desc inside /help also :)

done

mpiorowski (approved):

Issue #28, opened 2026-04-16
Accents not supported in chat input


LilDojd:
> [!NOTE]
> This issue was moved from #bugs room with Claude

## Description
Emoji are supported but accented characters are not rendered/accepted correctly in chat input.

mevanlc:
Seems to be working after the PR merge + deploy. Reopen if issues arise.

Issue #34, opened 2026-04-16
Backspace and alt+backspace broken — VTE parser gets stuck


LilDojd:
> [!NOTE]
> This issue was moved from #bugs room with Claude

## Description
`alt+backspace` (delete word) does not work. After attempting it, regular `backspace` also stops working. Root cause identified as VTE parser getting stuck.

mpiorowski:
Fixed, will be in the release.

PR #51, opened 2026-04-16
Support non-ASCII typing in TUI composers

### LINKED ISSUE #44: Can't type in cyrillic
## Description

When in the chat view, cannot type letters from Cyrillic alphabet. Copying and pasting them works

## Steps to Reproduce the Problem

1. Go into any room
1. Press `i` and try to write Cyrillic

## Expected Behavior

Text is written into a textarea:

"тест тест тест"

## Actual Behavior

No text is written

## Specifications

- Platform: macOS
- Connection method: ssh
- Terminal emulator: kitty

### LINKED ISSUE #50: FR: UTF-8 keyboard input support
## Description

Support non-ASCII UTF-8 input (Cyrillic, CJK, diacritics, etc.) across all text composers in the TUI. Related to issue #44.

### Is your feature request related to a problem?

Pasting works, but users whose primary language uses any non-ASCII alphabet cannot compose messages

Root cause: composer input handlers gate printable input on byte range `(32..127)`. Multi-byte UTF-8 sequences (all bytes ≥ 0x80) fall into the default arm and get discarded.

https://github.com/mpiorowski/late-sh/blob/62a782e500ae9df3f4ec7eb02bf98eb3ba988023/late-ssh/src/app/chat/input.rs#L45-L49

https://github.com/mpiorowski/late-sh/blob/62a782e500ae9df3f4ec7eb02bf98eb3ba988023/late-ssh/src/app/chat/news/input.rs#L19-L22

(profile composer shares the same pattern.)

`VtCollector::print` already receives a decoded `char` but re-encodes to bytes and emits one `ParsedInput::Byte` per byte, erasing char boundaries before the composer sees them:

https://github.com/mpiorowski/late-sh/blob/62a782e500ae9df3f4ec7eb02bf98eb3ba988023/late-ssh/src/app/input.rs#L143-L153

### What solution would you like?

Not sure, something to do with `ParsedInput::Char(char)`

But as a result: `тест`, `漢字`, `café`, `αβγ` type correctly. Autocomplete, bracketed paste, and mention matching continue to operate on `char`-level strings (already the case for paste).

### What alternatives have you considered?

N/A

### Do you have any additional context?

- Reproduction (from issue #44): open any room, press `i` to compose, try to type `тест`. Nothing appears. Pasting `тест` works.
- Affected call sites share the exact pattern `b if (32..127).contains(&b) => composer_push(b as char)`, so the fix is mechanical once the `Char` variant exists.

LilDojd:
Fixes #44. Closes #50 Typing Cyrillic, CJK, and other non-ASCII characters in the chat composer now works

What changed

- We don't ignore multi-byte UTF8, instead emit `ParsedInput::Char(char)`
- Chat, news, and profile composers accept unicode
- Added unit tests for the parser (Cyrillic/CJK/Latin-with-diacritics) + integration tests
- Drive-by: fixed strip_ansi test helper that mangled UTF-8 by walking bytes instead of chars.

Result

тест, 漢字, café, αβγ all type directly in any composer. Previous functionality not altered


LilDojd:
Also probably closes #28 if I correctly understood the original bug report

LilDojd:
Related to #34 
@mpiorowski I can also look into it if you would like

mpiorowski (changes requested):
Great one, but pls lets add it only to chat input :) Processing url and profile dont need it, and the profile service will reject it anyway ;p

mpiorowski:
And yeah, the #34 seems tricky, feel free to check inside

PR #53, opened 2026-04-17
fix(chat): prioritize online users in @ mention picker

### LINKED ISSUE #29: @ mention should prioritize online users
> [!NOTE]
> This issue was moved from #bugs room with Claude

## Description
Pressing `@` shows all users including offline ones. Online users should appear at the top; offline users at the bottom.

LilDojd:
Closes #29.

`@` mention picker now ranks online users first and dims offline rows so the useful names are where you look first.

## What changed

- **Ranking** -- sort candidates by `(online desc, name asc)`. Online set built once from `ActiveUsers`
- **Visual affordance**:
  - selected → `AMBER` + `BOLD` (unchanged)
  - online → `TEXT`
  - offline → `TEXT_FAINT`

## Result

In a room with a few online + many offline members, typing `@` now shows the online folks on top in bright text, with the offline roster greyed underneath. Cursor still moves through both groups so you can `@` an offline user when you want to.

## Screenshots

_Before:_

image

_After:_

image

## Tests

- Unit tests inline in `state.rs`:

## Repro with:

```fish
# Bump LATE_MAX_CONNS_PER_IP 
function lssh
    ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p 2222 $argv
end

for u in alice bob carol
    ssh-keygen -t ed25519 -f ~/.ssh/late_$u -N "" -C $u
    lssh -i ~/.ssh/late_$u $u@localhost   # first-login creates user row; `q` to quit
end


# Keep alice + bob connected, leave carol offline. In alice's composer: `c` → `i` → `@`.
# Expected: bob + alice on top, carol faint at bottom.
```

### DIFF
--- a/late-ssh/src/app/chat/state.rs
+++ b/late-ssh/src/app/chat/state.rs
@@ -830,6 +830,28 @@ impl ChatState {
         self.invalidate_composer_layout();
     }
 
+    pub fn composer_delete_right(&mut self) {
+        let char_count = self.composer.chars().count();
+        if self.composer_cursor >= char_count {
+            return;
+        }
+
+        let byte_pos = self
+            .composer
+            .char_indices()
+            .nth(self.composer_cursor)
+            .map(|(i, _)| i)
+            .unwrap_or(self.composer.len());
+        let next_byte = self
+            .composer
+            .char_indices()
+            .nth(self.composer_cursor + 1)
+            .map(|(i, _)| i)
+            .unwrap_or(self.composer.len());
+        self.composer.replace_range(byte_pos..next_byte, "");
+        self.invalidate_composer_layout();
+    }
+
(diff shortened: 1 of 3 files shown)


mpiorowski (changes requested):
Great one, just a few nitpikcs:

- rank_mention_matches calls to_ascii_lowercase() up to three times per candidate (prefix filter, online lookup, sort key). The sort comparator additionally allocates two Strings per comparison. On every keystroke while @-typing, for every candidate. Cheap to improve by lowercasing once into a (String, MentionMatch) tuple before sort, then returning just the MentionMatch.
- online_username_set builds the full online set on every keystroke even when the query already narrows to zero candidates.
- No test for online_username_set or the None branch

PR #54, opened 2026-04-17
Render typing-related screen updates immediately (up to 15ms ETR)

mevanlc:
** See discussion below **

ETR = effective tickrate; it means typing can "drive" the server to do a render-and-io-update pass at most once every 15ms. typing faster than that does not drive additional renders, rather the render updates will be delivered at the next 15ms window (or 66ms world tick, whichever arrives soonest). If keys are arriving at >15ms intervals, typing should feel instantaneous, except for any unavoidable network lag or CPU overloading from other tasks.


LilDojd:
Hi @mevanlc , thank you for your contribution! I think this is ultimately @mpiorowski decision to touch tickrate, but I have a few comments on the current implementation. The render loop for the session looks like this:

1. Loop body calls `render_once` every interval
2. `render_once` calls `app.tick()`
3. all other systems ticks are ticked in app.tick body :)

This means that if a global tickrate is changed dynamically, tickrates of bonsai, visualizer etc. change too. I think this is not ideal, especially if we add other systems that are coupled to a clock. I sent a message in `/dev` on how we can avoid this. The main idea is we separate render tick from "world" tick. This would unblock this change

But it is a huge change, so needs more planning. What do you think?

mpiorowski:
yeah, global faster rate can bite us in the ass. But this one gave me a simple idea, what if on every keypress we jsut trigger the render.

The simple way is to reuse our render loop, add some NOTIFY, so its either a loop or notify sginal:

```
• Probably not much if you keep it narrow: roughly 40-90 LOC.

  Most of it would be in late-ssh/src/ssh.rs:

  - add a Notify shared between input path and render task
  - change the render loop from pure interval.tick() to select! { interval.tick(), notify.notified() }
  - split render_once() into:
      - tick_and_render_once()
      - render_only_once()
        or add a boolean like advance_world: bool

  Then a smaller bit in the SSH input path:

  - after handling input bytes, call notify_one()

  If you also want paired-client messages or resize events to trigger immediate repaint, add another 10-30 LOC.

  So:

  - minimal “keystrokes trigger render”: ~40-60 LOC
  - more complete “all UI-visible events can wake render”: ~70-120 LOC

  The risky part is not LOC, it’s making sure concurrent render requests coalesce cleanly and you don’t accidentally call app.tick() on notify-driven renders.
```

LilDojd:
yeah notify-based is good

LilDojd:
But throttle the notify path!! Otherwise anyone can spam keystrokes and make rerenders too frequent. Smth like

```
const MIN_RENDER_GAP: Duration = Duration::from_millis(15);

// and in a loop somewhere..

 let (advance_world, should_quit) = tokio::select! {
      _ = world_tick.tick() => (true, false),
      _ = notify.notified() => {
          // Throttle input-driven renders to MIN_RENDER_GAP
          let since = last_render.elapsed();
          if since < MIN_RENDER_GAP {
              tokio::time::sleep(MIN_RENDER_GAP - since).await;
          }
          (false, false)
      }
  };

```

Or maybe even without sleep with `tokio_stream::StreamExt::throttle`

mevanlc:
> yeah, global faster rate can bite us in the ass. But this one gave me a simple idea, what if on every keypress we jsut trigger the render.

yes I also had the idea of keystrokes triggering a rerender in order to echo the typed keys ASAP. But, I still want to audit how to decouple world_state_tick from render_tick.  need to be able to call render_tick while being certain NONE of the world state is advancing more quickly than intended. since input into the composer is now "sticky", the keystroke-causes-immediate-render-tick idea is stronger now than before.  with sticky-compose-mode a lot more users will be idling in compose mode, making fast-tick much more expensive on CPU than it was before sticky-compose-mode.

my feeling is we should drop the unconditional speed-up-the-tickrate-when-in-compose-mode approach and rely entirely on keystrokes (in certain contexts) able to cause an ~immediate rerender, which some cap to how many re-renders can be keystroke-triggered, in order that someone can't dump a load of keystrokes on the server and eat more CPU than they should.

mpiorowski:
Ohh, now I see the problem, if we would only go with notify, then we would do the tick for an input, so wasteful resources.

I lvoe the idea of seperate render_once(), not connected with tick, and jsut the input wired to it. @mevanlc let dont go fancy now, everykeystokre with minimal throttle and see :)

mevanlc:
> Ohh, now I see the problem, if we would only go with notify, then we would do the tick for an input, so wasteful resources.
> 
> I lvoe the idea of seperate render_once(), not connected with tick, and jsut the input wired to it. @mevanlc let dont go fancy now, everykeystokre with minimal throttle and see :)

I'm prototyping it mostly as you said and LilDojd agreed.  So we'll have kind
of conceptually two things that can be "ticked":

1. world_update - animations and such - runs every ~66ms as before.  
2. io_update - recalculates TUI "visual" state and sends it to the user

world_update will continue to run on the 66ms schedule.  When it triggers it
also causes an io_update so that world state changes get pushed to the
connected client at least every 66ms, as before.  Conceptually, world_update
also causes an io_update.

io_update, because it is split out now, can be triggered by separate
conditions, not just world_update. The new condition we discussed here is:
user typing events.  User types a key -> server immediately wakes up, does
what it needs to do to handle the key stroke, does rerendering, and pushes
updated screen state to the client ASAP*.

*now because key events can trigger work on the server, we have to limit how
often we allow a key event to trigger that work, otherwise spamming keys could
trigger a lot of rerendering load and IO load on the server, leaving us
vulnerable to unintentional or intentional DOS on CPU or possibly even network.
If key events arrive faster than one per 15ms, we defer the extra render(s)
they would naively cause: the keystroke's effect on app state is applied
immediately (nothing lost), but the re-render is scheduled for the end of the
15ms throttle window. At that moment we render once, covering all accumulated
keystrokes. If a world tick happens to fire during the throttle window, it
absorbs the pending render and the throttle is discarded, since a world_update
also includes an io_update.

PR #56, opened 2026-04-17
Add bell notifications

Kyllingene:
Adds a setting in the profile to enable bell notifications.

This sends the ASCII BEL character (0x07) in addition to the OSC 9/OSC 777 notifications.


late-ssh/src/app/profile/ui.rs
```diff
@@ -234,33 +263,6 @@ fn build_lines<'a>(view: &ProfileRenderInput<'a>, width: u16) -> Vec<Line<'a>> {
         dim,
     )));
 
-    // ── Notifications ──
```
mpiorowski, line 267: why remove? :D

mpiorowski (changes requested):
Looks great, just one comment and no test for the toggle or for the 0x07 byte actually getting appended when notify_bell = true.

late-ssh/src/app/profile/ui.rs
Kyllingene: I think that snuck through when handling merge conflicts, sorry. I'll fix that.

Kyllingene:
As far as I can tell everything is in order. I refactored `desktop_notification_bytes` a little to make it easier to test the bell character getting added, and modified the tests in `late-ssh/app/render.rs` to match. I also added a check to make sure navigating to the bell checkbox and toggling it works as expected. `make check` passes with the new tests. Running locally, I can confirm the bell works as intended, and that the profile page + settings are working properly.

Let me know if there was anything else wrong!

mpiorowski:
Thank you!

Kyllingene:
Thank you!

PR #61, opened 2026-04-17
fix(vote): prevent duplicate activity notifications on revote

### LINKED ISSUE #37: Revoting same music type sends excessive notifications to all users
> [!NOTE]
> This issue was moved from #bugs room with Claude

## Description
Revoting for the same music type triggers a vote notification for every user each time, causing notification spam.

adi-sen:
## Summary
Fixes #37 - Revoting for the same music genre no longer sends excessive notifications to all users.

## Changes
- Modified `VoteService::cast_vote` to check the user's previous vote before upserting
- Only publishes activity events when the vote actually changes (different genre or first-time vote)
- Added integration test `does_not_emit_activity_when_revoting_same_genre` to verify the fix

## Test Plan
- [x] `cargo test -p late-ssh --lib vote::svc` (unit tests pass)
- [x] `cargo clippy -p late-ssh -- -D warnings` (no warnings)
- [x] `cargo fmt --check` (formatting correct)
- [ ] Integration test requires Docker/testcontainers (will run in CI)
- [ ] Manual: Vote for a genre, then vote for the same genre again - should not see duplicate activity notifications

## Behavior
**Before:** Every vote triggers an activity notification, even when voting for the same genre repeatedly.

**After:** Activity notifications only appear when:
- User votes for the first time
- User changes their vote to a different genre

Revoting the same genre updates the database timestamp but does not spam notifications.


mevanlc:
My main comment:

This only protects against activity spam if the user is making the same vote over and over again. I think maybe we should just completely mute a user's **voting** in the activity log after they cast a single vote -- for some cooldown period of time. So if someone is spamming vote keys in a cycle L, A, C, L, A, C, etc... they aren't able to clog the activity log and make it unfun for everyone by drowning out more serious voters.

For the PR code changes as they stand. Main finding:

1. Extra DB round-trip on every vote (svc.rs:202-205) - Every cast_vote now does a SELECT before the UPSERT. Cleaner and atomic: change Vote::upsert to return whether the genre changed in a single SQL statement.
 
Nitpicks:
1. svc.rs:220 comment "Only publish activity if the vote actually changed" restates what vote_changed already says — drop it.
1. svc.rs:202-205 and svc.rs:207-216 each take their own pooled connection. Not new to this patch but the two-checkout pattern is now doubled. The single-SQL fix collapses it.
1. Test (tests/vote/svc.rs): the 500ms timeouts will add ~500ms to the happy path -- is there a better way?

mpiorowski (approved):
Great PR, thank you!

PR #80, opened 2026-04-21
One word: Artboard.

mevanlc:
## Summary
This adds Artboard to the arcade: a shared multiplayer ASCII canvas wired into `late-ssh`, backed by the dartboard crates, with local state/input/rendering support and Postgres snapshot persistence.

The branch also rounds out the in-app Artboard workflow with selection and floating brush behavior, a glyph picker modal, a tabbed help overlay, and arcade hub updates so the feature reads as a first-class game instead of a hidden experiment.

## Why
The arcade had single-player games, but no shared real-time creative surface. Artboard fills that gap and makes the existing terminal UI/input work usable for collaborative drawing instead of treating the canvas as a special-case external component.

Persisting snapshots in Postgres also closes the biggest product gap in the earlier prototype: without persistence, the canvas state was too fragile for a shared feature that users are expected to return to.

## Impact
Users can open Artboard directly from the games hub, draw on a shared canvas, use richer editing controls, and recover the current board state across restarts.

From an implementation standpoint, this adds the Artboard game module, threads it through app state/render/input, introduces the snapshot storage model and migration, and adds test coverage around the new behavior.

## Validation
- `cargo test -p late-ssh`


mpiorowski (approved):
GOATTTTTT

PR #82, opened 2026-04-21
Fix order of help lines in overview section

MyGodItsFull0fStars:
Fixed the order of the primary screens in the overview.
Now they are aligned with the appearance in the app, and the keyboard "shortcuts" to navigate to each primary screen.


mevanlc:
lgtm

mpiorowski:
thank you ! :)

Issue #68, opened 2026-04-19
Audio buffering in wsl2


onliner94:
## Description

When using the cli-program inside a wsl, the continious audio signal starts to heavily buffer to the point, that the music is unhearable. 

## Steps to Reproduce the Problem

1. Using the cli-app inside a Ubuntu24.04 wsl2 environment (under Windows 10 with Windows Terminal)

## Expected Behavior
The audio starts normal. After 1-2 minutes a heavy buffering makes the audio unhearable.

## Actual Behavior
The audio starts normal. After 1-2 minutes a heavy buffering makes the audio unhearable.

## Specifications

- Platform: Ubuntu-24.04 (wsl2 on Windows 10)
- Connection method: cli-app
- Terminal emulator: Windows-Terminal

LilDojd:
Might be related: https://github.com/microsoft/wslg/issues/908

mpiorowski:
shoudl be fixed :)

PR #85, opened 2026-04-22
Cli audio playback against jittery sinks and flaky streams

LilDojd:
Addresses #68. Root cause on the reporter's setup is WSLg's PulseAudio-over-RDP bridge ([microsoft/wslg#908](https://github.com/microsoft/wslg/issues/908), #1429) upstream. We can't fix that, but we can stop the CLI from making it worse. The audio pipeline also had a few latent correctness/perf bugs worth clearing out while in here (cross-referenced with [pdeljanov/Symphonia#260](https://github.com/pdeljanov/Symphonia/discussions/260)).

- Lock-free SPSC queue. Replaced `Arc<Mutex<VecDeque<f32>>>` with `ringbuf::HeapRb`.
- Reuse `SampleBuffer` instead of frequently allocating
- Fixed cpal output buffer on Linux: `BufferSize::Fixed(2048)`

- [x] Ran make check


mpiorowski (changes requested):
I havent dive deep into review, but AI throw me a fast high:

High: late-cli/src/audio/decoder_thread.rs:89 now feeds the ring buffer with push_slice(pending), which can split the interleaved sample stream at any sample boundary. The consumer in late-cli/src/audio/output.rs:173 assumes it can pop a whole source_channels frame; when it gets only part of one, it drops the partial frame and outputs silence. On stereo this can turn L,R,L,R into L dropped, then R treated as the next frame’s left channel, so under pressure you get sample loss and channel skew. The old mutex queue avoided this by enqueuing whole chunks atomically. This needs frame-aligned writes or a consumer-side partial-frame stash.

mpiorowski:
fix on another pr :)

PR #92, opened 2026-04-23
Remove artboard push/pull

mevanlc:
The push/pull commands just mess up the artboard because nobody knows how to use them. And even people who do know how to use them (like me) don't find them very useful. Basically ppl are accidentally griefing the entire board with just a few keystrokes (accidental or not) for a lame feature nobody needs. I wanna remove it. :D

    - unmap the ctrl push/pull chords in artboard input handling
    - remove the old bindings from in-app help and context docs


mpiorowski (approved):
100% agree :D

PR #93, opened 2026-04-23
Native CLI local echo for chat composer + SSH input queue decoupling

mpiorowski:
## Summary
At ~60 concurrent SSH sessions we see noticeable input lag in the TUI — chat typing and screen switches feel sluggish because keystrokes contend on the same `TokioMutex<App>` that render holds for the full ratatui draw. This branch tackles that on two fronts: native `late-cli` now locally predicts the chat composer so paired users see keystrokes instantly regardless of server round-trip, and the SSH server's input path no longer blocks on the render mutex. It also adds a new admin `/fill-room` command that came along for the ride.

## What changed
- **Native CLI chat-composer local echo.** Server emits a `ChatComposerHint` over the existing `/api/ws/pair` control socket describing where the composer is drawn (x/y/width/height), its current text, and the cursor position. The CLI consumes those hints in `late-cli/src/predict.rs`, runs local keystrokes through a `vte` input parser, and paints an overlay directly to stdout so typed characters appear without waiting for the SSH frame. Scope is intentionally narrow: only the `Dashboard` chat card and the full `Chat` page, only native SSH mode, no mention popup, no artboard/games.
- **Prediction reconciliation.** When authoritative hints arrive, local predictions are kept if they still match exactly, dropped if the server moved in a direction we did not predict (submit/clear/external mutation), and preserved across duplicate hints that land before the WS round-trip catches up — covered by unit tests.
- **SSH input path no longer blocks on the render mutex.** `data()` now pushes raw bytes into a per-session `mpsc::UnboundedSender<Vec<u8>>` and the render task drains that queue under the mutex right before `tick()`/`render()`. An expensive frame can still delay paint, but it no longer stalls ingress for every other keystroke on that worker. Also enables `nodelay` and 30s/3x SSH keepalives on the server config.
- **Cursor styling cleanup.** The chat composer cursor uses explicit theme fg/bg pairs instead of the `REVERSED` modifier, so the local-echo overlay and server-side ratatui draw agree on cursor appearance.
- **New admin command `/fill-room #slug`.** Bulk-adds every user to a public room and flips `auto_join = true` for future joins. Rejects private rooms. Surfaced in the help modal alongside the existing `/create-room` and `/delete-room`.

## Behavior notes
- Local echo is best-effort and strictly additive: if the hint is missing, inactive, or the prediction diverges from an incoming authoritative hint, we simply stop overlaying and let server-side rendering drive the composer again.
- The SSH input queue is unbounded per session. Because the queue is drained every render tick (now woken within ~15ms via `RenderSignal`), backpressure in practice is bounded by the render cadence, not the channel.
- `/fill-room` is idempotent on already-joined users (add-all-users returns count of newly added) and enabling `auto_join` is a one-way door for the room until an admin edits it directly.

## Feature flags
None.

## Screenshots / Video
Still need to attach a before/after capture of typing into the Dashboard chat card on a laggy session — the effect is a visual/latency one and a static screenshot will not convey it. Flagging so reviewers can hold until video is attached.

## Testing
- **Automated**
  - `late-cli` — new `predict.rs` unit tests cover authoritative-reset dropping a stale prediction and duplicate hints preserving an in-flight prediction; existing `apply_pair_control_*` WS tests updated for the new `ChatComposerHint` variant.
  - `late-ssh` — `chat::state` tests for `parse_fill_room_command` (with/without `#`, empty, non-command) and `new_chat_textarea` cursor-style tests; integration tests `fill_room_task_adds_all_users_to_public_room` and `fill_room_task_rejects_private_room` in `late-ssh/tests/chat/svc.rs`.
  - Existing `ssh::tests::stale_permit_does_not_arm_throttle` still guards the render-signal invariant after the queue-drain refactor.
- **Manual**
  - Native `late-cli` → chat composer on both Dashboard and full Chat page: type a burst, confirm characters appear locally without waiting for the server frame; submit a message and confirm the overlay is cleared cleanly with no stale characters.
  - Repeat with `--ssh-mode old` to confirm the subprocess path is unaffected (no overlay, server-authoritative only).
  - Run `/fill-room #somepublicroom` as admin on a public room with a few existing members; confirm all users are added, `auto_join` is flipped, and the success banner reports the correct count. Try on a private room and confirm the "Only public rooms can be filled" error banner.
  - Drive ~60 concurrent SSH sessions (load harness) and confirm keystroke-to-paint latency is bounded by render cost rather than by mutex contention on the input side.


mevanlc:
Cool idea.  I'm curious to see if it encounters desync issues, though. Great if it works!

mevanlc:
@mpiorowski is it worth adding a --no-local-echo in case anyone has issues they can turn it off?

PR #94, opened 2026-04-23
Add live mentions counter to outer frame border (clickable)

mevanlc:
* add a right-aligned thing that lives in the top late.sh border 
* it tells you if you have unread mentions. 
* click it to jump to the mentions overview screen

CleanShot 2026-04-23 at 14 27 29@2x


mpiorowski (approved):
sweet

PR #103, opened 2026-04-24
Artboard: local paint color picker with Ctrl+U / Ctrl+Y

mpiorowski:
## Summary
Artboard painters were stuck with the single color assigned to them by the peer list, so they could not vary stroke color within a session. This PR adds a local 16-color palette that each painter can cycle through with `Ctrl+U` / `Ctrl+Y`, affecting their own typing, pastes, glyph-picker insertions, and floating previews without changing what peers see for them in the ownership overlay.

## What changed
- New `Ctrl+U` / `Ctrl+Y` keybindings cycle previous / next through a 16-entry paint palette defined in `artboard/state.rs`.
- The selected color drives subsequent typed glyphs, paste, glyph-picker insertion, swatch stamping, and the floating brush preview.
- The Artboard info sidebar now shows the active color (`Color  N/16  #RRGGBB`) plus a per-entry palette row with the active swatch marked.
- Palette state is per-user and local: it does not change the peer-assigned color used by the ownership overlay or peer list.
- Help overlay, Artboard help tab, and `CONTEXT.md` updated to document the new shortcut and what stays local.
- `Ctrl+U` / `Ctrl+Y` were previously listed as intentionally unbound push/pull chords; they are now explicitly reassigned to paint-color cycling.

## Behavior notes
- Cycling wraps at both ends (`rem_euclid` over 16 entries).
- Before the user cycles, the reported active index falls back to the palette slot matching the peer-assigned color, defaulting to index 1 when no match exists.
- Input handling routes `0x15` / `0x19` even when the help overlay is closed; the bytes are consumed so printable-glyph drawing is unaffected.

## Feature flags
- None.

## Screenshots / Video
- UI-visible change (new `Color` + `Palette` rows in the Artboard info sidebar). Screenshots / short capture of cycling should be attached before review.

## Testing
- Automated:
  - `ctrl_u_and_ctrl_y_cycle_paint_color` in `artboard/input.rs` verifies both keybindings are handled and advance the active index.
  - `paint_color_cycles_and_typed_glyphs_use_selection` in `artboard/state.rs` verifies a typed glyph picks up the cycled color on the canvas.
  - `paint_color_cycle_wraps` covers negative-delta wrap-around.
  - Existing sidebar snapshot tests updated to reflect the new `Color` and `Palette` lines.
- Manual:
  - Open Artboard, press `Ctrl+Y` / `Ctrl+U` and confirm the sidebar `Color` row and active palette marker update.
  - Type, paste, and insert from the glyph picker to confirm each uses the newly selected color.
  - Toggle the ownership overlay (`Ctrl+\`) and confirm peer colors / initials are unchanged by local cycling.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -3,7 +3,7 @@
 ## Metadata
 - Domain: late.sh - Terminal Clubhouse for Developers
 - Primary audience: LLM agents working on this codebase, human contributors
-- Last updated: 2026-04-24 (Artboard global help/gallery context)
+- Last updated: 2026-04-24 (Artboard paint color selector context)
 - Status: Active
 - Stability note: Sections marked `[STABLE]` should change rarely. Sections marked `[VOLATILE]` are expected to change often.
 
@@ -475,6 +475,7 @@ Only canvas mutations are shared. Editor affordances stay local to the current S
 - Active selection anchor
 - Floating brush / floating selection preview
 - Swatch strip contents + pin state
+- Selected paint color from the 16-color local palette (`Ctrl+U` / `Ctrl+Y`)
 - Temporary sampled glyph brush
 - Help overlay tab + scroll
 - Glyph picker search state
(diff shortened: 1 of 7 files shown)


mevanlc:
This may sound nitpicky, but forgive me I moonlight as a designer and stuff like this matters more to me than one might imagine. `:P`

The Info panel is getting way too crowded with stuff people don't need.  Since the Info panel takes up space over the canvas (basically, it gets in the way of viewing the canvas) we should take care to keep it small/minimal. Here's my suggestion:

CleanShot 2026-04-24 at 16 22 07@2x

***Edit: much improved mockup in separate comment below.***

and another way:

```
┌──────────────────────────────┐
│ Mode        active           │
│ Color      ┌────────────────┐│
│  ^Y / ^U   │                ││
│            └────────────────┘│
│ Cursor       x, y            │
│ Mouse        x, y            │
│ Owner        ....            │
│                              │
│ ─Users────────────────────── │
│                              │
│   user1                      │
│   user2                      │
│   userN...                   │
│                              │
└──────────────────────────────┘
```

Edit: actually I think 2wide 1high is a better presentation for the swatches because it reads more visually clearly:

image


## *ALSO: only 15 colors are visible to me, shouldn't there be 16?* 

- Yes I just checked the 16th (or 1st?) is white (or maybe "default") and not shown. Here's how I would incorporate it:

image

mevanlc (changes requested):
See my 1st comment in Conversation tab of PR.

mevanlc:
Better mockup:

PR #107, opened 2026-04-25
fix(ci): prevent shell injection via GitHub context interpolation

thereisnotime:
`${{ github.repository }}` and `${{ inputs.package }}` were interpolated directly into `run:` shell commands in `build.yml` and `ci.yml`. This is flagged by Semgrep rule `yaml.github-actions.security.run-shell-injection`.

If either value contains shell metacharacters, an attacker who controls the input (for example via a fork PR or a crafted workflow dispatch payload) could inject arbitrary commands into the runner and exfiltrate secrets.

The fix moves all context values into `env:` variables and references them as `$ENV_VAR` in the shell. GitHub's runner binds `env:` values as environment variables, so the shell receives them as data rather than code — no quoting bypass possible.

**build.yml** — `resolve_image_tag` step: `inputs.registry`, `github.repository`, `inputs.image_name`, `inputs.image_tag`
**ci.yml** — `cargo_clippy` and `cargo_test` steps: `inputs.package`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,8 +52,13 @@ jobs:
 
       - name: resolve_image_tag
         id: vars
+        env:
+          REGISTRY: ${{ inputs.registry }}
+          REPOSITORY: ${{ github.repository }}
+          IMAGE_NAME: ${{ inputs.image_name }}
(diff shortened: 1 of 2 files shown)


mpiorowski:
Thank you!

PR #114, opened 2026-04-27
Show reaction owners overlay when pressing f in reaction leader mode

mevanlc:
Press `f` again after pressing `f` to see the names of who reacted and with what reaction.


late-core/src/models/chat_message_reaction.rs
```diff
@@ -100,7 +101,8 @@ impl ChatMessageReaction {
             .query(
                 "SELECT message_id,
                         kind,
-                        COUNT(*)::bigint AS count
+                        COUNT(*)::bigint AS count,
+                        ARRAY_AGG(user_id ORDER BY created, user_id) AS user_ids
```
mpiorowski: Ok, this, for a 1000 messages with a lot of reactions, this will cost perf, we are wiring in on every snapshot change.

Great feature, but we could do it that we loads the reactions username on opening overlay only.

mpiorowski (changes requested):

late-core/src/models/chat_message_reaction.rs
mevanlc: > Ok, this, for a 1000 messages with a lot of reactions, this will cost perf, we are wiring in on every snapshot change.
> 
> Great feature, but we could do it that we loads the reactions username on opening overlay only.

You're right. Sending around the reaction details everywhere when only a few people will look at f+f infrequently is inefficient. Lazy load is the way to go. Kind of like /members I think?

PR #119, opened 2026-04-28
User preference: show settings on connect (or not)

mevanlc:
- New `show_settings_on_connect` profile setting (default on); when off, the settings modal no longer auto-opens on connect.

Users who fill out a bit of profile info get access to a settings section called "Special".  The new `show_settings_on_connect` setting lives there, along with a mysterious gem that no one seems to know the purpose of. 🤔


mpiorowski:
haha, love it :D

PR #122, opened 2026-04-28
Add Admin-Pinned Chat Messages to the Dashboard

mpiorowski:
## Summary
- Lets admins pin important chat messages so they remain visible above the normal dashboard chat stream.
- Keeps pinned message visibility scoped to room membership, preventing private or DM content from appearing for non-members.

## What changed
- Added admin-only `Ctrl+P` support for toggling the selected message’s pinned state.
- Displays pinned messages in a dedicated stack above dashboard chat when visible pinned messages exist.
- Includes pinned messages in chat snapshots and live pin update events.
- Adds database support for `chat_messages.pinned` with an index for pinned message lookup.

## Behavior notes
- Pinned messages are generic chat messages, not limited to announcements.
- Only admins can pin or unpin messages.
- Pinned messages are only sent to users who are members of the message’s room.
- Dashboard snapshots load up to 100 pinned messages visible to the current user.

## Feature flags
- None

## Screenshots / Video
- UI-visible change: pinned messages render above dashboard chat.
- Screenshots or video still need to be attached before review.

## Testing
- Automated:
  - Added coverage that admins can toggle a message pin.
  - Added coverage that non-admin users receive an admin-only failure when attempting to pin.
  - Updated dashboard/chat render tests for the new pinned message view state.
- Manual:
  - Not provided in the diff.


mevanlc:
Should we use atomic SQL? I've been trying to do it elsewhere:  Something like:

```
...
SET
    pinned = NOT pinned RETURNING *
```

PR #127, opened 2026-04-29
[draft] late-bastion: keep users connected across upgrades

### LINKED ISSUE #67: FR: allow users to stay connected across upgrades
My original issue text here got hilariously out of date with the progression of this discussion.  So I'm deleting it. To see the real deal info, read #127. Meantime, here's some of my revised sales pitch:

---

Each individual benefit might not seem like a lot, but I think it all adds up:

0) Priority number one: make staying connected to late.sh as seamless and easy as possible!
1) Users who, for whatever reason, idle through the 1hour k8s grace period will come back to their computer still connected (well, reconnected) to late.sh!
2) If we do get strong ephemeral state support, upgrade reconnects could *possibly* happen more automatically than having the user go through pressing `q` and `r`. But I still think `q`, `r` should seem at least a little friendlier than a full manual reconnect for most people, hopefully.
3) Although I do not address it in this PR, the groundwork for reassociating music automatically is here.
4) If for whatever reason we decide to keep late-cli on direct-ssh path, it can still benefit from some of the reconnect-awareness built here. But to be honest if we build the bastion correctly, I don't see any downside to routing everyone through it.
5) Having late-ssh able to accept connections over WebSocket lays interesting groundwork for spectator mode in a vanilla webbrowser. I am hesitant to let users in for fully-interactive via the web because the barrier to entry is too low and so we get bad eggs. But who can tell the future? I can't.
6) Although I stand by my suggestion to keep bastion as simple as possible (so it doesn't need upgrades), it does provide an interesting location to put BBS-style doorway experiences!  It is possible we could have a two-hop bastion system: bastion 1 is super-stable-don't-mess-with-it edition, bastion 2 has room for more complex, possibly less stable doorway experiences. Just a thought, NOT something I recommend, I just point it out as a possibility.
7) Possibility for some application-layer protocol-aware routing or late.sh "Fingerprint IdP" expansion in the future.
8) Bastion on separate hardware = SSH decryption load can be eaten by droplet VMs while late-ssh "big papa" processes get to speak cleartext websocket. Small optimization win maybe, but it's there.

mevanlc:
See:
* late-bastion/BASTION.md
* devdocs/LATE-CONNECTION-BASTION.md
* https://github.com/mpiorowski/late-sh/issues/67

---

### tl;dr
 
existing arch:
```
user ssh client <-ssh-> russh+late-ssh
                                  ^
                                  |
                             reboots for upgrades =
                             ssh connection dropped
                             user = :(
```

late-bastion arch:
```
                        russh+                tungstenite+ 
user ssh client <-ssh-> bastion  <-websocket->  late-ssh
                        ^                           ^                   
                        |                           |                  
           bastion is simple dumb           reboots for upgrades =  
           process rarely needs             websocket connection    
           upgrades = big uptime            dropped but bastion     
                                            doesn't care it keeps   
                                            ssh alive and keeps     
                                            retrying late-ssh until 
                                            it is back up, user     
                                            reconnects automatically
                                            user = :)               
```

***Note that, as a bonus, the websocket endpoint on late-ssh happens to lay substantial groundwork towards allowing users with vanilla web browsers (xterm.js) to connect to late-ssh -- perhaps for view only, or perhaps even to participate. Neat, huh? 🚀***

* Also there is the possibility of automatically resuming music, see https://raw.githubusercontent.com/mpiorowski/late-sh/f2dc23594637cc37291054d39e32bb84d4bf8fff/devdocs/IDEAS-BASTION-AND-MUSIC.md


mpiorowski:
@mevanlc the work itself is great, want to be clear on that. blocker is product, not code.

with the bastion the ssh tunnel survives a deploy but the App resets, someone mid-hand at a poker table blinks back to the dashboard with their bet lost. the current path at least shows the draining banner so they can cash out first.

i want to make the pods ephemeral first so state survives the swap. once that's done your bastion drops in cleanly, none of this work is wasted, just stacked on the wrong foundation right now.

parking for later. happy to sync on the ephemeral-pod plan whenever.

mevanlc:
> @mevanlc the work itself is great, want to be clear on that. blocker is product, not code.
> 
> with the bastion the ssh tunnel survives a deploy but the App resets, someone mid-hand at a poker table blinks back to the dashboard with their bet lost. the current path at least shows the draining banner so they can cash out first.
> 
> i want to make the pods ephemeral first so state survives the swap. once that's done your bastion drops in cleanly, none of this work is wasted, just stacked on the wrong foundation right now.
> 
> parking for later. happy to sync on the ephemeral-pod plan whenever.

@mpiorowski 

Aren't persistent connections purely an upgrade in user experience from what we have today? There is no user experience downgrade (or change) necessary because of this PR. For example the Bastion does not prevent implementation of a draining / cashout UX on Blackjack.

Maybe I misunderstand how long a user is allowed to stay connected to a prior version after an upgrade is deployed. I thought I noticed a significant drop in # of users connected after upgrades?

mpiorowski:
Yeah, they have 1h to drop ;p

And the active users you see, thats for the NEW pod. Only :D so you are able to chat with a person that is not in the active list, cos hes sitting in the "old" server.

mevanlc:
> Yeah, they have 1h to drop ;p
> 
> And the active users you see, thats for the NEW pod. Only :D so you are able to chat with a person that is not in the active list, cos hes sitting in the "old" server.

It can still work that way with bastion -- and without difficulty.

Let me try to motivate this PR a bit more:

- Bastion is not just an architectural thought-piece, it has real implications for (A) usability and (B) our ability to attract new users (C) and retain existing users.

- Bastion's backend changes opens the way for people to observe late.sh via browser via a real, functioning terminal (we could think of it as a "livedemo" mode to hook people?  Think of it like sales-screenshots on steroids.)

- Although it would need to be carefully considered, Bastion is a place that we could put cross-service or Doorway-type experiences.

Let me push back on your objections:

#### C1: Bastion does not fix cross-upgrade persistence
It isn't meant to fix that. But it is far from pointless. Also, it does not get in the way of state-preservation fixes, either.

#### C2: Bastion prevents users from having a graceful upgrade path (1hr to "cash out" and reconnect.)
It does not get in the way of having this UX at all. Core late-sh still retains complete control over when to disconnect users, and it has control over whether the bastion enters reconnection mode or simply allows the user to drop, as today.

---

We should consider having the active user count total up all users across old-version and new-version; number of connected users is a selling point for people ("social credit") -- people like being where other people are. Seeing a big number on active users readout makes people feel like "this is a place other people like to hang out, so I should keep hanging out here too."

mevanlc:
Sorry for the wall of text, but I wanted to make something clear: in terms of UX ***bastion is a strict superset of current functionality.*** It also doesn't get in the way of future improvements like enhanced state-preservation.  Already in this PR without tweaks we can recreate all of today's user experience exactly. But bastion opens the door to so much more than that.

---
                                                                                                                                                                                                            
 The UX around reconnect is still just as under late-core (aka late-ssh) backend's control as it ever was.  Today's upgrade experience consists of:                                                         
                                                                                                                                                                                                            
 1) In-app banner notifying "An upgrade is available. Reconnect to see it now." (Bastion does not change this.)                                                                                             
 2) It's worth noting that the user may or may not see the banner (alt-tabbed away, possibly for long periods of time, perhaps just listening to late.sh's music); either way, user may or may not choose   
 to reconnect on their own manually by using their ssh client software.                                                                                                                                     
 3) Users which choose not to reconnect (or the 1hr time chooses for them) are dumped out to their ssh client without special explanation.                                                                  
                                                                                                                                                                                                            
---                                                                                                                                                                                   
 
 Let's look at that user experience as it might look with bastion in play:
                                                                                                                                                                                   
1) Late-core can still request bastion do this exact disconnect flow. It's already under control of the websocket disconnect codes.  Additionally it's trivially possible to send even richer instructions via a late-core -> bastion disconnect JSON payload.                                                                                                          
 2) Bastion does not need to auto-reconnect and is already designed to be told what to do around this. All of these possibilities are trivially implemented                                                 
     - Bastion could leave them at a bastion-rendered prompt that says something like: "upgrade happened, press enter to reconnect or ctrl c to disconnect"                                                 
     - Bastion could say nothing at all reconnecting user silently                                                                                                                                          
     - Bastion could disconnect/dump user with exact UX as exists today.                                                                                                                                    
 3) Post-upgrade late-core is free to raise a customized explanation/welcome screen when user is reconnected by the bastion. It has (or trivially can have) all this information:                           
     - Is it the first connection this user has done to an upgraded instance?
     - What was the version of the user's last-connected instance?                                                  
     - Whether the prior connection was through the bastion and whether their new connection is through the bastion                                                                                         
     - Whether the user was reconnected by the bastion automatically or the current connection was user-initiated                                                                                           
     So the "confused user silently teleported to a reset-state" has multiple avenues of redress -- none of which are difficult to implement

mpiorowski:
Yeah, valid points. The usability win is real, but honestly not that big a deal today, and people who'd leave over a forced reconnect probably weren't sticking anyway. The livedemo angle is probably the strongest argument here.

C1 I concede: bastion is orthogonal to persistence work and does not block it.

C2 is where I still disagree.  Yes, bastion can support graceful drain in theory. But this PR does not keep that behavior today. When late-ssh drains, it sends close 1000. Bastion treats that as “reconnect”, so the user can be moved into a fresh app without choosing it.

To keep today’s behavior, late-ssh would need to say “do not reconnect, end this session” or “show a prompt first”. That policy is not in this PR.

And if we add that policy, the benefit becomes smaller. The user no longer runs `ssh late.sh` again, but they still have to press Enter or reconnect manually. That is nicer, but it is still an interruption. It does not solve the real problem: the app state is gone.

So I still want to park it for now. What I can promise you is that I will start working on the ephemeral state as soon as possible :)

mevanlc:
To address your concerns about reconnect UX and user-story, I'll make a push to this PR to address that more intentionally and head-on.

For the rest... I'm just gonna share my chat with Claude here because I feel it's the most honest way to reply:

Me: 
> being honest, the main reason I buckled down and implemented the bastion was because I disliked seeing the user count drop off whenever we push an upgrade to add a theme or tweak reactions or something. Those things ultimately are not nearly as substantial as preserving and growing the community. And for community we need connected users.  We've been upgrading at least once a day and I feel like we're kind of kicking people off and creating a friction point to stickiness/user adoption when it would be so much nicer to be moving the needle in the opposite direction. In a world where moving the "Checkout" button 15 pixels to the right increases webshop sales, making staying connected a little more frictionless may not be as small as it sounds like instinctually. To say "they weren't going to stay anyway if they don't reconnect manually" and that's probably true of ~95% of users. But that 5% compounds week to week into a number that's worth caring about. It's a statistics game and going by gut feel on one of the most important parts of the project - active community - is kind of a bummer. The way I motivated this PR:  doorways, webdemo, easier reconnect "just because" maybe wasn't the right way to frame this up: the ⚡️ that got me going was keeping the user count up and I genuinely think this can help with that.
> 
> Another thing is, you can externalize ephemeral state all day long (great idea btw!) but you're going to break reality before you're able to externalize a live ssh-encrypted TCP socket. The socket lives and dies with an instance, and as long as there are things to upgrade in that instance, the connection will drop during upgrades. If late.sh is kind of like a social web site but over the ssh-client everyone has built in, then bastion is our WAF or reverse proxy. It's a key part of web design, so why not make it a part of our ssh-acting-like-a-social-website arch? :)

Claude (paraphrased):
> "They weren't sticking anyway" reminds me of a conversion-funnel fallacy: the same shape as "people who didn't buy weren't going to buy anyway." So many A/B tests of micro-friction in the last 20 years (checkout button placement, form field count, password requirements, "are you sure?" prompts) has shown that gut-feel about user resilience is often wrong. People aren't rational about consumption cost; they're satisficers. A friction point at exactly the moment they're not actively engaged ("I alt-tabbed for a while while listening to the music, ate lunch, came back, terminal says disconnected") is perhaps not the best stickiness.
> 
> You also have a multiplier he might not be weighing: daily deploys. Even a 5% per-deploy drop compounds - assuming they stay gone, that's 30%+ of your active-at-deploy-time cohort lost per week. That's maybe worst-case, but what if it isn't?

mpiorowski:
Ahhh, you rly make me rethink it :D You sneaky one :D

Can you try to push the working "with banner" reconnect here? I think its about handling code 4000, not 1000. And let me sit on it one more time :)

mevanlc:
Ah, okay. So, a lot was missing from the original PR! Good to be skeptical.

Here's the new story:

1) Users connected via direct SSH should not see any UX difference if this PR merges.

2) Users connected via-Bastion get the same UX as direct SSH when they quit a non-draining instance: `q`, `q`.

3) The via-Bastion experience differs when the backend connection drops unexpectedly, for example a late-ssh crash, network break, or the old instance finally exiting after the 1h k8s grace period. In that case, bastion retries on the user’s behalf for up to 30s.
   1. If reconnect succeeds, the user sees the normal late.sh welcome sequence plus a 6s toast saying they were reconnected due to an upgrade or network problem. 
   2. If reconnect does not succeed within 30s, bastion drops their SSH connection.

4) The main via-Bastion difference is during upgrade draining. 
   1. Via-Bastion users see an “Update available” banner telling them they can press `q`, then **`r`**, to connect to the updated late.sh. Their `q` dialog differs slightly also: it reminds the `r` key is available. Also `q` is still supported there to do a real disconnect. 
   2. If they choose `r`, bastion retries on their behalf for up to 30s. On success, late-ssh receives a reconnect-reason header and shows a friendlier 6s “Welcome to the updated late.sh” toast.
   3. Note that I did fix an oversight/bug where via-Bastion (ws:// /tunnel) connections were not keeping the instance alive for the 1hr k8s grace period.

mpiorowski:
Will try to review it over weekend, thx for beeing considered

mpiorowski:
Ok, omg I had a LOT of thinking about it....and I was just about to hit the merge btn, cos everything works....but I just couldnt...

The main reason is that when I looked one last time at the whole picture....yes, we have a better ux, user dont need to re prompt `ssh late.sh` or `late`.

But there is cost, and is real and permanent. 7k LOC isn't the half of it. Were adding:
- one more thing on every keystroke. ssh -> bastion -> ws -> late-ssh. forever
- if bastion dies, everyone dies at once. v1 is single replica so its literally a SPOF for ssh
- "thin bastion" stops being thin the moment cli goes through it. exec, tokens, identity. then bastion knows users and we have two late-ssh
-  every weird bug now has 3 places to look. ssh client? bastion? late-ssh? two log streams to correlate
- cutover itself is weeks of careful work. dual run, watch, rollback. all on me

Thats the bill, every week, forever, for as long as the bastion lives. 

If we are really worried that this may cause some of our users to leave us, ship ephemeral pods, than add metrics, count it, check. Then make a decision.

So yes,sorry to say, but I still want to park it.

BUT I've heard you about the web live page, which is a great idea :) So what I did, extracted the tunnel part, and make it working ;)

https://late.sh/play

mevanlc:
I have to respect that it's your project and your decision. We discussed and you decided. So for me it's really okay to park it or just close it out. 

---

Extra credit only. ** Don't read ** if you're tired of thinking about this item. Because I want to see the best for late.sh and I really believe some arch like this is important and inevitable for growth. I can't help but reply more and share why I think so.

1. I think better UX (and thus possibly better retention) is worth 7k lines. It's not just about `q`, `r`, it's also about users who want to idle with late.sh in the background. They will ride out the upgrade timer and then get disconnected without something like this.  "Ambient" users matter, too!

It wouldn't be acceptable for Discord to exit to desktop every time they push a backend upgrade and I feel the same is similar here. But a BIG part of late.sh's charm is being able to access it from a vanilla ssh client. Unlike Discord who control a client that can retry/reconnect, the only way to keep vanilla-ssh users connected across upgrades is to decouple the connection from the "business logic" or "world."  There are a limited number of ways to do this decoupling and IMO they all look something like bastion. Single-point-of-failure is addressable by spinning up 1 bastion per connection, or even having it be a sidecar component that lives and dies with its late-ssh.

2. It does NOT have to be a forever commitment.  This change is easily backed out.  The old arch and a bastion-scrubbed codebase is just 1 reverse merge and redeploy away.  AI is great at reverse merges. Seriously easy, it doesn't need to be a big commitment.

3. I don't see how CLI presents a big problem. Admittedly I haven't looked into it deeply. But forwarding ssh sideband stuff over websocket is pretty darn easy.  As long as the world is in something like late-ssh, I don't think it's fair to say there are two late-sh.

4. Yes it's definitely extra complexity. There is almost never a new feature that reduces complexity.
But bastion, while a non-trivial amount of code, is conceptually very simple and thus easier to troubleshoot than it might seem. If there is even trouble. There might be none.

But ultimately my argument is simple: disconnecting users when you don't have to disconnect them is bad for business and something like a bastion (== decouple ssh connection from backend) is the only way to solve this dissatisfier for vanilla-ssh users. Ephemeral pods can't solve the vanilla ssh client limitation, sadly.

mpiorowski:
Ive increassde the draining pod life to 6 h, lets hope this will prevent the "supirse" disconnect for users ;p

mpiorowski:
closing for now :)

PR #130, opened 2026-04-29
Prototype browser (xterm+websocket) access to late-ssh (observe only)

mevanlc:
The websocket-based protocol that enables the bastion to keep late-ssh users connected across backend restarts can also be extended to support accessing the late-ssh experience through a web browser.  Start with observe-only/view only. Later PRs could investigate interactivity.


mpiorowski:
already done ;p

PR #147, opened 2026-05-02
[draft] add account [self-]deletion flow

### LINKED ISSUE #97: FR: Allow users to delete their own account
Allow users to delete their own account.  Common courtesy and standard practice for services with "accounts."

(We know it's "theirs" because they must be authenticated to access the function.)

mevanlc:
## Summary

- Adds an Account tab in Settings with a Delete Account button and username super-confirmation dialog.
- Deletes the current account through the profile service, terminates active sessions, and ends the deleting user's current app session.
- Adds in-memory deleting tracking during account deletion to reject reconnects and new session-token creation while deletion is in progress.
- Preserves moderation actor history and server bans when related user rows are deleted.

## Why

Deleting an account previously had no user-facing flow. The new confirmation dialog makes the destructive action explicit by requiring the current username before deletion proceeds.

The in-memory deleting tracking is specifically there to prevent a reconnect race: after a user successfully confirms deletion, they could otherwise reconnect or establish another connection just before the database delete commits. That could leave a session connected for a user whose account is deleted moments later. The deletion gate marks the account/fingerprint as deleting before async DB work, blocks new SSH auth/session-token creation while held, terminates active sessions, and only then clears.

For moderation, actor UUIDs and server-ban target UUIDs may now dangle intentionally. This keeps audit history and fingerprint/IP bans in place even when the actor or banned target account is deleted.

## Validation

- `cargo check -p late-ssh`
- `cargo test -p late-ssh delete_account_preserves_moderation_rows_and_allows_key_reuse`
- `cargo test -p late-ssh account_delete_confirmation_rejects_wrong_username_in_dialog`
- `cargo test -p late-ssh delete_account_preserves_server_ban_against_deleted_target`
- `cargo test -p late-core migrations_apply_and_track`
- Pre-commit rustfmt and clippy passed during commit.
- Stood up a `late-test` Docker compose instance and verified SSH/API/web/Postgres health.


mevanlc:
There's larger issue that this brings up which is syncing state and notifying state changes (both in-memory and db) between concurrently-running instances.  The in memory "deleting" gate guards against in-flight connections from getting past the "user exists" check and the next instant the account is deleted, but user stays connected. But only in the 1 late-ssh instance where the memory guard exists. We could move the deleting gate to the db which is stronger since it's shared state but there's still the question of how to notify other instances to re-check state due to something important happening (like user delete, or maybe server ban). This is a general problem of externalizing state and especially async state and avoiding data/state races where it's important to avoid them.

mpiorowski:
omg, I need to finally sit to this one :D

PR #148, opened 2026-05-02
 Default to IPv4-only for late.sh connection

eapotapov:
# Default to IPv4-only for late.sh connections

Default to IPv4-only across all three SSH modes. Owned by a single new
module, `late-cli/src/ipv4.rs`:

- **Native (russh)**: `connect_ipv4_only` resolves the target via
  `getaddrinfo`, drops v6 candidates, and hands a connected `TcpStream`
  to `russh::client::connect_stream` (replaces the previous
  `client::connect((host, port), …)` call).

- **OpenSSH and `old` (subprocess) modes**: `apply_ipv4_default` prepends
  `-4` to the ssh argv at config-load time. All five places that consume
  `config.ssh_bin` (the four `*_command_spec` builders and
  `spawn_subprocess_ssh`) automatically pick it up — none of them need
  to know about address families.

A user-supplied `--ssh-bin "ssh -6"` still wins via OpenSSH's
last-flag-wins ordering for address-family flags, so an explicit
override is supported without an escape hatch flag in this PR.

## Diff shape

| File | Change | Lines |
|---|---|---|
| `late-cli/src/ipv4.rs` *(new)* | Owns both helpers, module docstring explains the policy and both failure modes, three inline tests for the pure-logic helper | +118 |
| `late-cli/src/main.rs` | `mod ipv4;` | +1 |
| `late-cli/src/config.rs` | Single call to `ipv4::apply_ipv4_default(...)` at end of `Config::from_args` | +1 |
| `late-cli/src/ssh.rs` | Replace one `client::connect` call with `connect_ipv4_only` + `connect_stream` | +2 -1 |

Existing-code touch is **5 lines across 3 files**, all at well-defined
seams.

## Test plan

- [x] `cargo fmt -p late-cli --check`
- [x] `cargo clippy -p late-cli --all-targets -- -D warnings`
- [x] `cargo test -p late-cli --bin late` — 39 passed (3 new + 36 existing)
- [x] `make check` — full workspace, 921 tests, 0 failed, 0 skipped
- [x] Manual on a broken-IPv6 macOS network: `late --ssh-mode openssh`
      and `late --ssh-mode old` now connect immediately instead of
      hanging or returning "Connection closed by …" — `-4` is in the
      ssh argv. `late` (native) connects via the resolved A record and
      never attempts AAAA.

## Inline tests

`apply_ipv4_default` covers all three modes and the empty-input edge:

- `prepends_minus_four_in_openssh_mode`
- `prepends_minus_four_in_subprocess_mode`
- `leaves_native_mode_untouched`

`connect_ipv4_only` is not unit-tested — exercising it requires real
network resolution, which would be a network-dependent flaky test
rather than meaningful coverage.

### DIFF
--- a/late-cli/src/config.rs
+++ b/late-cli/src/config.rs
@@ -96,6 +96,8 @@ impl Config {
             }
         }
 
+        let ssh_bin = crate::ipv4::apply_ipv4_default(ssh_bin, ssh_mode);
+
         Ok(Self {
             ssh_target,
             ssh_port,
--- a/late-cli/src/ipv4.rs
+++ b/late-cli/src/ipv4.rs
@@ -1,0 +1,113 @@
+//! IPv4-only connection strategy for the `late` CLI.
+//!
+//! late.sh users whose network advertises IPv6 but cannot reliably use it
+//! hit two distinct failures with the default address-family behavior of
+//! both russh and `ssh(1)`:
+//!
+//! 1. **Silent v6 blackhole.** No SYN-ACK ever comes back from the v6
+//!    address. OpenSSH waits out its connect timeout (~75 s) before
+//!    falling back to v4, leaving the user staring at a hung terminal.
+//!
+//! 2. **v6 TCP works but the SSH session is closed immediately.** The
+//!    user sees `Connection closed by <v6-addr> port 22` within
+//!    milliseconds — something in front of the v6 endpoint (firewall,
+//!    fail2ban, ban list, misconfigured listener) accepted the TCP
+//!    handshake and then reset or hung up before banner exchange.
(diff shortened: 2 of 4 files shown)


eapotapov:
note: PR includes https://github.com/mpiorowski/late-sh/pull/146 change

one additional issue appeared during the testing - it looks like native connection has issues on a server side when late cli is running on slow connection (I'm using mobile network vpn for tests), I'll create an issue for this

mpiorowski:
than kyou for the fix, but i just went to other way and fucking finally properly allowed ipv6 :) Could you test it? @eapotapov

PR #151, opened 2026-05-03
Fix web pairing URL

Crestwave:
The API URL was updated to include `https://` in commit 4e3eeb65406289ebc1f309af684e671e45074958, but the web player was not updated to strip it out when connecting to the websocket. This results in a `NS_ERROR_UNKNOWN_HOST` error as it tries to connect to `wss://https//api.late.sh/api/ws/pair?token=secret`.

This patch simply strips out the added prefix. Tested successfully using an overrides script on the live website, as the bug does not apply to localhost.

image

Note: this might need to be applied to `chat/page.html` as well. I did not include it in this PR as I could not tell if it is still in use or not.
https://github.com/mpiorowski/late-sh/blob/2bfd5703f4f07c95ea3a9c2f048d69bfe032fe15/late-web/src/pages/chat/page.html#L107


mpiorowski:
goat fix, thank you

PR #171, opened 2026-05-07
Make spacebar room jump keys case sensitive, add A-Z

mevanlc:
The number of rooms now exceeds 36  (a-z 0-9).
Add uppercase A-Z as additional shortcuts.
Required making jump keys case sensitive.


mpiorowski:
@mevanlc !! was just thinking about this one.... what if we go the other way? similar to have nvim does it, realitves letters. 

image

mpiorowski:
Or a combination of yours plus this one really

mpiorowski:
what do you think?

mevanlc:
> Or a combination of yours plus this one really

Combination seems good. We already have case sensitive shortcuts elsewhere so it's not a new idea.  But also relative you never run out?

mpiorowski:
You wanna add it? :D

mpiorowski:
Ive decided to add only yours....the bakcweards started feel strange when i tested it

PR #185, opened 2026-05-11
allow right sidebar visibility to be set per screen #

mevanlc:
I personally like to have it turned on for all screens except off on 2 and off on 5.

Maybe other people would appreciate deeper customizability here?

---

    - persist on/off/custom sidebar mode and per-screen visibility
    - add settings modal controls for custom sidebar screens
    - honor custom sidebar visibility in rendering and input layout


mpiorowski:
goat, but the newest udpate fucked it hard :D

mevanlc:
@mpiorowski just finished another merge from main

PR #209, opened 2026-05-19
feat: birthday field, tracking, /birthday command + connect alert

hardlygospel:

---

Contributed by **Tasmania** (Tony Hosaroygard — [hardlygospel.github.io](https://hardlygospel.github.io)) — with thanks to Mateusz Piórowski for late.sh.

## Summary

- Adds a year-less `MM-DD` **birthday** to the user `settings` JSONB (privacy: no year), via the same `extract_*` pattern as bio/country.
- **`/birthday` chat command** to set it: `/birthday 03-07` sets (accepts `M-D`, `MM-DD`, `MM/DD`), `/birthday` shows the current value, `/birthday clear` removes it.
- Shows the birthday on the **profile** card as e.g. `🎂 Birthday: 7 March — in 3 days` / `— today!`.
- Adds **one-way birthday tracking**: press **`t`** on someone's profile to track/untrack their birthday. Stored as `tracked_user_ids` in settings, persisted through the existing `ProfileParams` path exactly like `favorite_room_ids`. Not a mutual friendship — the tracked user isn't notified and it grants nothing.
- **Connect-time birthday alert** delivered through late.sh's existing profile event → banner notification pipeline (the same `ProfileEvent` mechanism that surfaces "Profile saved!"). On connect, tracked users whose birthday is today or within the next 7 days are summarised in one banner, e.g. `🎂 ada — birthday today! · bo's birthday in 3 days`.
- New pure `models::birthday` module (parse/validate `MM-DD`, days-until-next, is-today, is-upcoming, `month_day_label` for "7 March" formatting) and a pure `build_birthday_alert` summariser — both fully unit-tested incl. Feb-29-on-non-leap-year, year-wrap, today-first ordering, and the 7-day window edge.

## Notes

- **Setter is a chat command, not a settings-modal row.** The settings modal renders ~30 rows into a fixed-index layout; adding an edit row there is fragile and wants a visual test. A `/birthday` command keeps the setter fully self-contained — it reuses the existing `requested_*` submit pattern (`submit_composer` validates via `normalize_birthday` and stashes a `BirthdayRequest`; `handle_post_submit_requests` applies it via `ProfileState::set_birthday`, which persists through the existing `ProfileParams` path). A modal row remains a clean follow-up for anyone who wants it.
- **Notification mechanism:** the in-app `notifications` table is mention/DM-specific (NOT NULL `actor_id`/`message_id`/`room_id`, no `kind` column), so a birthday can't be a row there without an invasive schema change. The alert instead rides the existing `ProfileEvent` → banner system, which is the natural fit and needs no schema/migration. Promoting it to an OSC desktop notification (gated by a new `birthdays` entry in `notify_kinds`) is a clean follow-up if you'd like that.
- **Test-target fix:** the latest `main` merge introduced four `ProfileParams { ... }` initializers in `tests/profile` and `tests/chat` without this PR's new `birthday` / `tracked_user_ids` fields — the test target did not compile. This PR fills those in.

## Test plan

- [x] `cargo fmt --all --check` / `cargo clippy -p late-core -p late-ssh --all-targets -D warnings` clean
- [x] `models::birthday` unit tests (8, incl. `month_day_label`) + `profile::svc` birthday-alert tests (4) + `/birthday` parse tests (3) pass
- [ ] Reviewer: `/birthday 03-07` sets and confirms with a 🎂 banner; `/birthday` echoes it; `/birthday clear` removes it; birthday renders on profile; `t` toggles tracking and persists across reconnect; connect banner appears when a tracked user's birthday is today/within a week


mpiorowski (changes requested):
There is no ui to put the birthday :D

hardlygospel:
Added the missing setter. The feature previously had display, tracking, and the connect banner but no way for a user to actually set their own birthday — so the field stayed empty end-to-end. New chat command:

- `/birthday 03-07` — set it (accepts `M-D`, `MM-DD`, `MM/DD`)
- `/birthday` — show your current birthday
- `/birthday clear` — remove it

Went with a chat command rather than a settings-modal row deliberately: it stays self-contained and avoids surgery on the modal's fixed-index layout. Parsing reuses the existing `requested_*` pattern — `submit_composer` validates via `normalize_birthday` and stashes a `BirthdayRequest`, `handle_post_submit_requests` applies it through a new `ProfileState::set_birthday` (persists via the existing `ProfileParams` path).

Also in this commit:
- New pure `birthday::month_day_label` ("03-07" → "7 March"); the profile modal's `format_birthday` now uses it and drops its duplicate month table.
- 🎂 on the profile birthday line.
- Fixed four `ProfileParams` initializers in `tests/` that the latest `main` merge introduced without the `birthday` / `tracked_user_ids` fields — the test target did not compile before this.

hardlygospel:
> There is no ui to put the birthday :D

This has been stepped back into a / command and resubmitted.

hardlygospel:
Rebased onto current `main`. Three conflict files, all resolved as parallel-additions alongside the now-merged `/petname` plumbing (no logic conflicts — both commands wire through identical `requested_*` / `take_requested_*` patterns):

- `late-ssh/src/app/profile_modal/ui.rs` — birthday row sits above the new chips row in the personal-info block
- `late-ssh/src/app/chat/state.rs` — `requested_petname` and `requested_birthday` coexist; both parse blocks live side-by-side in `submit_composer`; tests for both kept
- `late-ssh/src/app/chat/input.rs` — `apply_petname_request` and `apply_birthday_request` as separate fns in the slash-command dispatcher

Local verification was constrained this round (disk pressure on my MacBook), so I'm leaning on CI for the full gate rather than re-running `make check` locally. The diff against the previous push is pure conflict-marker removal — happy to rebase again if anything breaks.

PR #219, opened 2026-05-21
feat: terminal-sourced color theme

tealblu:
hey all, this is a little larger than just a theme, but hopefully it's ok with you! - indi (fawn on late.sh ^-^)

- theme name: Terminal
- theme id: `terminal`
- visual direction: sources colors from terminal-provided text colors. intended to be used alongside a theming system like matugen, pywal, etc
- accessibility considerations: entirely and wholly depends on the end user to have a terminal which supports this feature and to use a sensible theme. screenshot colors were generated by Matugen, using material design principles. end-user experience is not guaranteed to be coherent.
- screenshots:
image
image
image


mpiorowski:
Thank you!

PR #221, opened 2026-05-21
feat: welcome-back banner with last-seen + gentle reminders

hardlygospel:
## Summary

Adds a small ambient banner shown when a user connects to late.sh after at least an hour away (or for the first time). The message lands in the existing toast slot at top-right and reads, for example:

> ✿ welcome back, @hardlygospel — last on 3 hours ago. remember to be kind 💛

For brand-new users:

> ✿ welcome to late.sh, @hardlygospel — pull up a comfy pillow and chill 🛋

The reminder rotates day-to-day (seeded by `user_id` + day-of-year), and the banner is **suppressed when the previous session was less than an hour ago** so quick reconnects stay quiet.

## What's in it

- **New module `late_core::welcome`** — pure helpers: `pick_reminder(seed)`, `humanize_since(then, now)`, `build_welcome_message(...)`. No clock or DB dependency; all logic unit-tested with explicit `now` parameter.
- **New `BannerKind::Welcome`** — theme `AMBER`, icon `✿`, 15 s TTL (vs. 5 s for success/error) so the line is readable before it fades.
- **`ensure_user` / `ensure_web_tunnel_user`** now also return the pre-update `last_seen` value, so the welcome line can quote it before it's overwritten by `update_last_seen`.
- **`SessionConfig` / `SessionBootstrapInputs`** gain a `previous_last_seen: Option<DateTime<Utc>>` field; `App::new` builds the banner once at startup and stashes it on `self.banner` so the existing toast pipeline takes care of rendering/expiry.
- **No schema changes.** Only existing `users.last_seen` is read.
- **8 reminders** in the initial list (bonsai, comfy pillow, kindness, stretch, water, soft with someone, look out a window, no rush to reply). `pick_reminder` is modular so reordering or extending the list is safe — no contract on order.

## Tests

`cargo test -p late-core` covers:
- bucket boundaries of `humanize_since` (sub-minute → "a moment ago", up through years)
- defensive handling of future-dated timestamps
- determinism + modular wrap of `pick_reminder`
- non-empty + unique reminder list
- `build_welcome_message` suppresses quick reconnects, returns first-visit greeting for `None`, and includes the relative-time string for returning users

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test -p late-core --lib` (89 passed, including 9 new welcome tests)
- [ ] Manual: SSH in, disconnect, wait >1h, reconnect → banner shows with relative time
- [ ] Manual: SSH in, disconnect, reconnect within 1h → banner suppressed
- [ ] Manual: brand-new user → first-visit greeting


mpiorowski:
A little bit too much code for the gains, sorry. 

Every new LOC = new code to maintain.

PR #223, opened 2026-05-21
feat: rotate farewell messages on disconnect

hardlygospel:
## Summary

Replaces the single hardcoded \"Stay late. Code safe. ✨\" exit string with a small pool picked from at random on each disconnect. Same warmth, a bit more variety. The original line stays in the rotation as a familiar default.

Companion piece to #221 (welcome banner on the way in) — this one's on the way out.

## What's in it

- **New module `late-ssh::app::common::farewell`**: 10-line pool, pure `pick(seed)`, runtime `pick_random()` (via `OsRng`, same entropy source `splash_tips` uses), `render_exit_payload()` which wraps the choice in `\r\n` for the SSH/web-tunnel `data` calls.
- **SSH (`late-ssh/src/ssh.rs`)** and **web-tunnel (`late-ssh/src/web_tunnel.rs`)** exit handlers both call `render_exit_payload()`, so the two paths stay in sync without duplicating the string.
- Removed the two `const EXIT_MESSAGE: &str = ...` declarations.

## Pool

> Stay late. Code safe. ✨
> Sleep well. The bonsai is in good hands. 🌱
> Until next time. ☕
> Catch you on the next late shift.
> We'll keep the lights on. 🕯
> The cat says goodbye. 🐈
> Take it easy out there.
> Mind the witching hour.
> Don't forget to drink water.
> Goodnight, friend.

Easy to add or swap any of these — the pool is just a `&[&str]` and `pick` is modular.

## Test plan

- [x] \`cargo fmt --check\`
- [x] \`cargo clippy -p late-ssh --all-targets -- -D warnings\`
- [x] \`cargo test -p late-ssh --lib farewell\` — 6 passed
- [ ] Manual: SSH in and \`exit\` a few times; confirm different lines appear

### DIFF
--- a/late-ssh/src/app/common/farewell.rs
+++ b/late-ssh/src/app/common/farewell.rs
@@ -1,0 +1,99 @@
+//! Exit-banner content shown to the user when their session ends. A small
+//! pool that's picked from at random; the original "Stay late. Code safe."
+//! stays in the rotation as a familiar default.
+
+use rand_core::{OsRng, RngCore};
+
+/// Pool of farewell messages printed on disconnect. Order is not part of any
+/// contract; `pick` always selects via modulo so reordering or extending the
+/// list is safe.
+pub const FAREWELLS: &[&str] = &[
+    "Stay late. Code safe. ✨",
+    "Sleep well. The bonsai is in good hands. 🌱",
+    "Until next time. ☕",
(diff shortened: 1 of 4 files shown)


mpiorowski:
Same, a little too much LOC to make it worth ;p

PR #227, opened 2026-05-22
feat: collapsible room-list sections

hardlygospel:

---

## Summary

Each room-rail section header (favorites / core / channels / updates / dms) gets a leading toggle — `-` expanded, `+` collapsed. Click the header row to fold or unfold that section.

- Everything starts **expanded** — default behaviour is unchanged for anyone who never clicks a header.
- Collapse state is **session-only**: it resets to all-expanded on reconnect. No schema change, no settings plumbing.
- The toggle glyph renders in `theme::TEXT_BRIGHT` so it stays legible against every theme background.

Mainly aimed at the `channels` section, which can run long.

## Implementation

It rides the existing `visual_order_for_rooms` chokepoint. That one function produces the ordered `Vec<RoomSlot>` that *every* navigation path reads — `j`/`k`, `^N`/`^P`, the number jumps — and the renderer reads it too. Dropping a collapsed section's rooms there means navigation and rendering both skip them with **no changes to any navigation handler**. The renderer still emits the header for a collapsed section so it can be reopened.

- New `RoomSection` enum + `ChatState::collapsed_sections: HashSet<RoomSection>`.
- `toggle_section` snaps the selection to the first still-visible slot if a collapse would otherwise strand the cursor inside a hidden section.
- `room_list_section_hit_test` maps a header-row click to its section; it's checked before `room_list_hit_test` so header clicks toggle instead of selecting a room.
- `favorite_ids` is now derived from the raw favorite list rather than the (collapse-filtered) visual order, so collapsing Favorites can't leak favorites into Core/Channels.

Toggling is **mouse-only** in this version. A keyboard fold key is a small follow-up — happy to add it here if you'd prefer it land together.

## Test plan

- [x] `cargo fmt --check` / `cargo clippy -p late-ssh --all-targets -- -D warnings` clean
- [x] `cargo test -p late-ssh --lib -- app::chat` — 130 passed, incl. new tests: collapsed sections drop their rooms from `visual_order_for_rooms` (one per section), Discover stays (not collapsible), `RoomSection` label round-trip
- [ ] Reviewer: click each header to collapse/expand; collapse the section holding the active room and confirm the selection snaps somewhere visible; confirm the `+`/`-` is readable across a few themes


mpiorowski:
GOAT

PR #233, opened 2026-05-23
Add Asterion room

ricott1:
A first sketch of adding [Asterion](https://github.com/ricott1/asterion). Still need to finalize a PR there https://github.com/ricott1/asterion/pull/7, but local tests are encouraging.


ricott1:
@mpiorowski I took the late.sh approach of implementing the RoomGameManager trait and treating Asterion as other table games like Poker. Not sure if this is the approach u had in mind, another one would be a simple proxy to my game servers, but I saw u dont have this kind of stuff in the codebase rn, so I went for the full late.sh embedding approach. I split Asterion into a core crate so that the UI is basically all on late.sh side

ricott1:
@mpiorowski Im not sure what should be the rpeferred way to drop/close a room. Should we close it if inactive for X minutes, for instance?

PR #234, opened 2026-05-23
feat: ambient aquarium band under every screen

hardlygospel:

---

## Summary

You mentioned wanting to work out how to have the fish-tank/aquarium — floating, every screen, full width, "doesn't have to be perfect". This is a v1 take.

- Reserves a **2-row water band at the bottom of every screen**, full terminal width — drawn beneath Dashboard, Arcade, Rooms and Artboard alike.
- Top row: slowly drifting ripples (`~` / `-`).
- Bottom row: a small shoal of fish swimming across — two species (`><>` common, `><(('>` long), facing whichever way they're going.
- **Per-user**: species, colour, direction, speed and starting position are seeded deterministically from `user_id`. Each user sees their own stable aquarium across sessions, with **no new DB state** needed for v1.
- Modals still render into `inner`, so they overlay the band while open — same way toast banners already behave. The band is "under it all" positionally; modals are temporary.

If we later want the persistent customisation from the old PLAN.md `aquarium_state` shape (tank tiers, decor, owned fish species from the marketplace), that drops in on top of this without changing the renderer — just swap "seeded from user_id" for "loaded from DB".

## Scope

| | |
|---|---|
| Files | 3 new (`app/aquarium/{mod,state,ui}.rs`) + 4 touched (`app/{mod,state,tick,render}.rs`) |
| Diff | +297 / -2 |
| New deps | none |
| New DB migrations | none |

## Test plan

- [x] `cargo build -p late-ssh` clean
- [x] `cargo test -p late-ssh --lib` — 830 tests pass, including 6 new aquarium tests
- [x] `cargo clippy -p late-ssh --all-targets -- -D warnings` exits 0
- [x] Render-buffer test (`TestBackend`) confirms wave + fish glyphs actually paint across full width, and fish positions advance between ticks
- [x] Determinism test confirms same `user_id` → same shoal; different users → different shoals
- [ ] Visual smoke-test on a real terminal

## Notes

- Band height (`WATER_HEIGHT`) is a single const if you want it 1 row or 3.
- Fish glyphs are kept ASCII-only (no emoji) to avoid width-mismatch issues across terminals.
- Picks up automatically on every screen because the reservation lives in the outer `draw()` in `render.rs` — no per-screen wiring.

### DIFF
--- a/late-ssh/src/app/aquarium/mod.rs
+++ b/late-ssh/src/app/aquarium/mod.rs
@@ -1,0 +1,2 @@
+pub mod state;
+pub mod ui;
--- a/late-ssh/src/app/aquarium/state.rs
+++ b/late-ssh/src/app/aquarium/state.rs
@@ -1,0 +1,138 @@
+use ratatui::style::Color;
+use uuid::Uuid;
+
+pub const WATER_HEIGHT: u16 = 2;
+pub const POS_SCALE: i64 = 1000;
+
+const FISH_MIN: usize = 5;
+const FISH_MAX: usize = 8;
+const SPEED_MIN: i32 = 60;
+const SPEED_MAX: i32 = 200;
+
+#[derive(Clone, Copy, Debug)]
+pub enum Species {
+    Common,
+    Long,
+}
+
+#[derive(Clone, Debug)]
+pub struct Fish {
+    pub x_milli: i64,
+    pub speed_milli: i32,
+    pub species: Species,
+    pub colour: Color,
+}
+
+#[derive(Clone, Debug)]
+pub struct WaterState {
+    pub fish: Vec<Fish>,
+    pub ticks: u64,
+}
+
+impl WaterState {
+    pub fn new_for_user(user_id: Uuid) -> Self {
+        let mut rng = Xor64::from_uuid(user_id);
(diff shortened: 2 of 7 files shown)


mpiorowski:
yes, sth like that, but I am w8 for a proper shop to add the aquarium :)

hardlygospel:
Rebased onto current `main` — clean compile, 6 aquarium tests still pass, `cargo clippy -p late-ssh --lib -- -D warnings` clean.

One thing worth raising: main now has a **separate** aquarium feature — the `show_aquarium_modal` / `app/hub/aquarium/` modal (admin-mod only). This PR's module sits at `app/aquarium/` and does the ambient band on every screen — different feature, different module path, both compile happily, but the shared word is confusing for a reader.

Three options, your call:

1. **Land as-is** — two "aquarium" features at different paths, distinguished by the modal-vs-band difference.
2. **Rename my module** to `app/water_band/` (or similar — `ambient_water`, `aquarium_band`, whatever fits) so it's obviously a different thing from the hub modal.
3. **Close this PR** if you'd rather extend the hub aquarium into the ambient role instead.

Happy with any of the above. If you'd like a rename, point me at the name you'd prefer and I'll force-push.

mpiorowski:
added :)

PR #247, opened 2026-05-26
feat: daily care streak for cat companion

hardlygospel:
Thanks Mat — small follow-up in the same lane as #240, same "feels more alive per LOC" lever on the cat.

---

## Summary

Single per-user care streak: any care action (feed, water, play, groom, treat) keeps the streak alive once per day. Surfaced on the cat modal mood line when the streak reaches 2+ days:

```
content  .  mostly cared for  .  7-day streak
```

That's the entire user-visible change.

## Why

Cheap "perceived-change-per-LOC" extension on top of #205. Honours the **"neither pet ever dies"** rule: a lapsed streak silently sits at 0 — no shaming banner, no neglect downgrade, just stops counting until the next care day brings it back to 1.

Daily granularity (not per-action) is intentional — you can't grind the counter by spamming feed/water/play in the same session.

## Scope

| | |
|---|---|
| Files | `late-core/migrations/059_add_cat_care_streak.sql` (+3), `late-core/src/models/cat.rs` (+29 / -37 — five `touch_*` methods consolidated through a shared `touch_care` helper), `late-core/tests/cat.rs` (+93), `late-ssh/src/app/cat/state.rs` (+72), `late-ssh/src/app/cat/modal_ui.rs` (+13) |
| Diff | +210 / -37 |
| Migrations | `059_add_cat_care_streak.sql` — adds `care_streak_days INTEGER NOT NULL DEFAULT 0` + `care_streak_last_day DATE` to `cat_companions` |
| New deps | none |

Streak math lives in SQL on each `touch_care` update so concurrent care actions can't race the counter:

```sql
care_streak_days = CASE
  WHEN care_streak_last_day = current_date     THEN care_streak_days
  WHEN care_streak_last_day = current_date - 1 THEN care_streak_days + 1
  ELSE 1
END
```

`CatState::bump_streak_today` mirrors the same logic locally so the modal reflects the new count immediately, before the spawned `feed_task` / `water_task` / `play_task` lands. `CatState::care_streak()` returns 0 once the streak has lapsed (last care day older than yesterday) so a stale "42-day streak" never lingers in the UI.

## Test plan

Pure-logic unit tests inline (per CONTRIBUTING):

- `display_streak_shows_streak_when_cared_for_today`
- `display_streak_keeps_streak_when_yesterday_was_last_care`
- `display_streak_lapses_when_older_than_yesterday`
- `display_streak_zero_when_no_care_recorded`

Integration tests in `late-core/tests/cat.rs` against a testcontainers Postgres:

- `new_companion_starts_with_zero_streak`
- `first_care_action_sets_streak_to_one`
- `multiple_care_actions_same_day_keep_streak_at_one`
- `care_after_yesterday_extends_streak`
- `care_after_a_gap_resets_streak_to_one`

### DIFF
--- a/late-core/migrations/059_add_cat_care_streak.sql
+++ b/late-core/migrations/059_add_cat_care_streak.sql
@@ -1,0 +1,3 @@
+ALTER TABLE cat_companions
+    ADD COLUMN care_streak_days INTEGER NOT NULL DEFAULT 0,
+    ADD COLUMN care_streak_last_day DATE;
--- a/late-core/src/models/cat.rs
+++ b/late-core/src/models/cat.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, NaiveDate, Utc};
 use tokio_postgres::Client;
 use uuid::Uuid;
 
@@ -16,6 +16,8 @@ crate::user_scoped_model! {
         pub last_groomed: Option<DateTime<Utc>>,
         pub last_treated: Option<DateTime<Utc>>,
         pub name: Option<String>,
+        pub care_streak_days: i32,
+        pub care_streak_last_day: Option<NaiveDate>,
     }
 }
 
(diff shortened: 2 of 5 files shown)


mpiorowski:
A little bit too much LOC for the effect

PR #254, opened 2026-05-26
fix: wipe Sixel pixels before next frame on transitions

hardlygospel:
Thanks Mat — small terminal-graphics fix. The image modal leaves Sixel pixels behind on WezTerm (and likely other Sixel-mode terminals); resize is the only way to clear them. Reported in-app: \"Sixel not appearing to clear on wezterm; when opening another image, it gets put behind the image that didn't get cleared.\"

---

## Summary

Sixel has no delete-by-id protocol like Kitty does, and on some terminals the Sixel raster layer persists above cell content until the cells underneath are written to. So closing the image modal leaves the image on screen, and opening a new image composes it under the stale one.

Fix is **pre-frame**, written directly to \`self.shared\` so the wipe bytes land in the output buffer **ahead of** ratatui's frame diff. ratatui's normal cell writes then overwrite the wiped area with the correct new content.

- \`TerminalImageRenderState::pre_frame_sixel_wipe_bytes\` returns spaces (\`\\x1b[0m\` + cursor-position + per-row space fill) over the previous frame's Sixel placements. Only fires on transitions — image modal closed, image swapped, foreground overlay opened — so a steady image-modal frame is a no-op (no bandwidth penalty).
- \`build_commands\` now takes \`suppress_sixel\`. While a foreground overlay (\`show_settings\`, \`icon_picker_open\`, hub, help, terminal-help, mod, profile, bonsai, splash, quit-confirm, web-chat-qr, pair, room-search, booth) is open, Sixel re-emission is skipped so the image doesn't repaint on top of overlay cells. When the overlay closes, the state transition triggers a fresh re-emit.

Kitty's existing delete-by-id cleanup path is untouched.

## Scope

| | |
|---|---|
| Files | 2 touched (\`app/files/terminal_image.rs\`, \`app/render.rs\`) + 1 doc (\`CONTEXT.md\`) |
| Diff | +237 / −5 |
| New deps | none |
| New DB migrations | none |

## Test plan

- [x] New unit tests for: skip-when-unchanged, fire-on-modal-close, fire-on-image-swap, fire-on-overlay-open, no-op-without-prior-Sixel, suppress-Sixel-under-overlay
- [ ] On WezTerm in Sixel mode: open image modal, close it — image disappears cleanly
- [ ] Open image modal, open another image — new image replaces the old, no compositing
- [ ] Image modal open, open icon picker (\`Ctrl+]\`) — Sixel disappears under the picker; closing picker brings it back
- [ ] Steady image-modal frame doesn't re-emit Sixel data each tick (check bandwidth doesn't spike)
- [ ] Kitty terminals (Kitty / Ghostty / WezTerm in Kitty mode) unaffected

## Notes

The fix only does work when the previous frame had a Sixel placement, so terminals that never use Sixel (Kitty / iTerm2 / no-protocol sessions) hit zero overhead. WezTerm should normally land on Kitty graphics — if you see Sixel on WezTerm, \`echo \$TERM \$TERM_PROGRAM\` will show why detection didn't pick the Kitty path.


hardlygospel:
Example of bug : https://files.late.sh/chat/2026/05/019e64c7-c348-7380-b2a8-628b4cf967b9.png

Astavie:
Oh! I submitted the bug report assuming that it was using Sixel, but WezTerm indeed also supports the kitty image protocol so I'm now unsure what it actually used ?

maybe there's a problem clearing kitty images on wezterm :shrug: 

image

Astavie:
yeah: wezterm uses kitty, but it doesn't delete images because late deletes images by z-index, but wezterm does not support this as of yet (relevant wezterm code: https://github.com/wezterm/wezterm/blob/577474d89ee61aef4a48145cdec82a638d874751/term/src/terminalstate/kitty.rs#L241)
an easy fix is to maybe move wezterm to either iterm2 or sixel, as it also (claims to) support these protocols

mpiorowski:
Went with @Astavie suggestion

PR #260, opened 2026-05-27
feat: /coffee and /tea commands post a rotating ASCII cup

hardlygospel:
Thanks Mat — small chat ritual in the \"feels alive per LOC\" lane. Same flavour as the cat/bonsai polish but on the chat side.

---

## Summary

\`/coffee\` and \`/tea\` post a short ASCII cup to the current room as a coffee/tea-break ritual. No arguments, just the command.

```
  )  )
 ( ( (
  c[_]
```

- Coffee uses the mug-with-handle silhouette (\`c[_]\`); tea uses the handle-less cup (\`\\___/\`). Steam patterns are shared and rotate through four variants via \`ChatState::next_cup_variant\` (session-local counter, not persisted) so rapid back-to-back rituals don't look identical.
- Routes through the existing \`send_message_with_reply_task\` send path so the message is a regular chat message subject to the same length/visibility rules. No new chat-kind, no new DB.
- Parser sits before \`unknown_slash_command\` so typos like \`/coffe\` still get the \"Unknown command\" banner.

## Scope

| | |
|---|---|
| Files | 1 touched (\`app/chat/state.rs\`) + 1 doc (\`app/chat/CONTEXT.md\`) |
| Diff | +122 / −0 |
| New deps | none |
| New DB migrations | none |

## Test plan

- [x] New unit tests cover: case-insensitive matching, argument/typo rejection, kind-specific silhouettes, variant rotation wrapping at \`CUP_VARIANT_COUNT\`
- [ ] Type \`/coffee\` in \`#lounge\` — posts a coffee cup as a normal message
- [ ] Type \`/tea\` — posts a tea cup (different silhouette)
- [ ] Run \`/coffee\` four times — see four different steam patterns, fifth wraps back to first
- [ ] Type \`/coffee please\` — falls through to \"Unknown command\" banner (rejects arguments)


mpiorowski (changes requested):
have you tested it ? :D cos its not wroking

hardlygospel:
Thanks for catching that Mat — I had \`clear_composer_after_submit()\` running before reading \`composer_room_id\`, and that clear nulls the room. So \`let Some(room_id) = self.composer_room_id\` always bailed and the send never fired. Embarrassing — apologies.

Fixed in 8c833a7: snapshot the room id before the clear, then send. \`cargo test\` and the four existing unit tests still pass (they cover the parser + steam-variant rotation but not the actual send path, which is the bit that was broken). The path the fix takes is the same one every other \`send_message_with_reply_task\` call uses, so I'm confident it'll work end-to-end now — let me know if it still misbehaves and I'll spin up the docker stack here to verify in-app.

PR #261, opened 2026-05-27
feat: ambient 'here Xh · ♫ Yh' row under the sidebar clock

hardlygospel:
Thanks Mat — picking up another item from CONTEXT.md §7 \"ambient presence\" (listed there alongside quiet hours / typing indicator). Tiny diff, same \"feels alive per LOC\" lane.

---

## Summary

Adds one dim line directly under the sidebar clock that reads:

- \`here 2h 14m\` as soon as you arrive.
- \`here 2h 14m  ·  ♫ 1h\` once a browser/CLI is paired up for audio.

Glanceable session ambience, no settings to toggle, no DB or service.

- \`format_short_duration\` in \`app/common/time.rs\` picks the unit: \`47s\` / \`14m\` / \`2h\` / \`2h 14m\` / \`1d 4h\`. Caps at days so multi-day SSH sessions don't render \`76h 14m\`.
- New \`session_started_at: Instant\` on \`App\` (set in \`App::new\`) and \`audio_paired_at: Option<Instant>\` (set/cleared in \`App::render\` based on \`paired_client_state()\`).
- The presence row borrows the visualizer's drop priority — kept while music + cat + bonsai still have headroom for one more line, dropped first on tight sidebars. No rule between the clock and presence; they read as one stacked header.
- \`audio_paired_at\` flips back to \`None\` when pairing drops, so a fresh browser pair-up restarts the music timer rather than resuming the old one. Session timer never resets within a session — reconnects spawn a new \`App\`.

## Scope

| | |
|---|---|
| Files | 4 touched (\`app/state.rs\`, \`app/render.rs\`, \`app/common/{sidebar,time}.rs\`) |
| Diff | +137 / −2 |
| New deps | none |
| New DB migrations | none |

## Test plan

- [x] New unit tests for \`format_short_duration\` covering all five buckets (seconds, minutes, hours, hours+minutes, days+hours)
- [ ] Sidebar shows \`here Xs/Xm/Xh\` under the clock; updates as the session ages
- [ ] Pair audio via browser or CLI — \`· ♫ Xs\` segment appears
- [ ] Disconnect audio — music tail disappears, here-timer keeps counting
- [ ] Reconnect audio — music timer starts from zero, not the previous value
- [ ] Shrink the SSH window vertically — presence row drops with the visualizer; clock + music/cat/bonsai survive

### DIFF
--- a/late-ssh/src/app/common/sidebar.rs
+++ b/late-ssh/src/app/common/sidebar.rs
@@ -34,6 +34,9 @@ const MUSIC_QUEUE_HEIGHT: u16 = 2;
 const BONSAI_MIN_HEIGHT: u16 = 16;
 // Cat: 3 art rows + 1 footer row.
 const CAT_HEIGHT: u16 = 4;
+// Presence row: 1 line below the clock, no rule. Same drop priority as
+// the visualizer — keeps the sidebar usable on short windows.
+const PRESENCE_HEIGHT: u16 = 1;
 
 pub struct SidebarProps<'a> {
     pub game_selection: usize,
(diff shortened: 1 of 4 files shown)


mpiorowski:
Not worth 1 row of space, I know it doesnt sound much, but every row counts on smaller screen, sorry

PR #277, opened 2026-05-29
feat: sender-side pixel cup celebration + /sixelcheck diagnostic

hardlygospel:
Thanks Mat — two small additions in the same sixel/chat-ritual lane as the wipe fix (#254) and the cup commands (#260). Purely additive on top of the existing terminal-image pipeline.

---

## Summary

Two related sixel features, bundled because they share the same plumbing.

### 1. Sender-side pixel cup celebration

When the user sends \`/coffee\` or \`/tea\`, the chat ritual itself still goes out to every viewer as plain ASCII (unchanged from today). On the sender's own terminal, if Sixel / Kitty / iTerm2 images are detected, a small pixel mug appears in the bottom-right corner of the chat area for ~3 seconds — a private treat for the act of brewing.

- Cup is drawn procedurally with \`image::RgbaImage\` (no bundled assets), PNG-encoded once per \`(kind, protocol)\`, cached for the lifetime of the process.
- Routed through the existing \`terminal_image::terminal_image_from_bytes\` pipeline that already backs the chat image modal — same wipe semantics, same protocol gating, same fallback shape ("no image protocol → nothing extra renders").
- Receivers see only the existing ASCII chat message. No new broadcast, no new DB, no new persistent UI surface — the cup placement is just a transient overlay on the sender's render frame.

### 2. \`/sixelcheck\` diagnostic command

One-shot slash command (same lane as \`/coffee\`, \`/tea\`, \`/brb\`, \`/petname\`) that prints a banner naming the detected terminal-image protocol. Local banner only — does not post to chat.

Output shapes:
- \`Terminal images: Sixel ✓ (this is the protocol used for the cup ritual)\`
- \`Terminal images: Kitty graphics ✓\`
- \`Terminal images: iTerm2 inline images ✓\`
- \`Terminal images: not detected. Try a Sixel/Kitty/iTerm2-capable terminal (foot, Kitty, Ghostty, WezTerm, iTerm2 …).\`
- \`Terminal images: disabled for this session (tmux passthrough off or override env set).\`

Useful for users on weird terminals who are wondering why images look the way they do, and for our own bug-hunting next time anything sixel-shaped misbehaves.

## Scope

| | |
|---|---|
| Files | 5 modified + 1 new (\`chat/cup_sixel.rs\`) |
| Diff | +447 / −3 |
| New deps | none |
| New DB migrations | none |

## Implementation notes

- \`terminal_image_from_bytes\` was previously module-private; bumped to \`pub(crate)\` so the cup module can encode without going through the URL-fetch path. No functional change.
- Cup module owns a \`OnceLock<Mutex<HashMap>>\` cache so repeated rituals don't re-encode the image.
- Placement uses a stable per-kind UUID so consecutive frames within the 3s window are treated as the same placement and the renderer skips redundant byte emission (same dedupe path the chat image modal uses).
- Cup expiry runs in \`ChatState::tick()\` next to the other tick drains, so no separate timer.

## Test plan

- [x] \`cargo test -p late-ssh --lib -- chat::state::tests chat::cup_sixel\` — 11 tests pass (4 new cup_sixel tests covering cache hit, display size, sixel-only-for-sixel, non-trivial pixel coverage; 2 \`/sixelcheck\` parser tests; 1 \`CupCelebration\` expiry test; existing 4 cup parser/art tests still green)
- [ ] Manual smoke: run \`/coffee\` in WezTerm / Kitty / foot — see the chat ritual *and* a small pixel mug in the corner
- [ ] Manual smoke: run \`/coffee\` in macOS Terminal or generic xterm — see only the chat ritual, no errors
- [ ] Manual smoke: run \`/sixelcheck\` across terminals and confirm the banner reports the right thing


mpiorowski:
A little to much LOC for the action ;p forgive

PR #278, opened 2026-05-29
feat: pixel trophies for top 3 + pixel shop unlock celebration

hardlygospel:
Thanks Mat — companion drop to #277 in the same sixel-polish lane. Two small visual additions on the Hub, both opaque on top of the existing terminal-image pipeline. No new persistent UI, no new shop categories, no new stat tracking — just pixel polish on existing surfaces.

---

### 1. Leaderboard top-3 trophies

The Top Chips and Arcade Wins panels in the Leaderboard tab get tiny gold / silver / bronze pixel trophies overlaid on the leading prefix cells of ranks 1, 2, 3. Score panels (Tetris / 2048 / Snake) are left alone — they're denser and already have monthly + all-time pairs per game.

- Trophies render only on capable terminals (Sixel / Kitty / iTerm2). Anywhere else the rows look exactly as they do today — the trophy slot is just the existing prefix (\`   \` / \` › \`).
- Trophies only appear for ranks that actually exist in the entries list, so an empty board never gets a stranded gold-cup sitting over the \"no chip earnings yet this month\" placeholder.
- 3 cells wide × 1 cell tall, anchored to the leftmost column of each row.

### 2. Shop purchase celebration

The Hub Shop body gets a brief pixel-sparkle burst centered over the item list after a successful purchase. Lasts ~3 seconds, then disappears.

- **Only first-time unlocks trigger the burst.** Equip / unequip / aquarium-fish-quantity adjustments are deliberately not celebrated.
- This is gated by a \`pending_purchase\` marker set in \`ShopState::activate_selected\` (in the purchase branches) and consumed when the matching \`ShopEvent::ActionCompleted\` arrives. \`ActionFailed\` clears the marker so a later unrelated success doesn't get celebrated as if it were the purchase.
- Banner text is unchanged — the burst is the visual companion to the existing \"Purchased X\" success banner. On terminals without image support, only the banner is visible.

## Implementation notes

- Both sprites are drawn procedurally with \`image::RgbaImage\` (no bundled PNG assets), PNG-encoded once per cache key, and routed through the shared \`terminal_image::terminal_image_from_bytes\` pipeline — same wipe semantics (#254), same protocol gating, same fallback shape as the cup ritual from #277.
- Trophy slots and the celebration rect are stored on \`HubState\` via \`Cell\` and populated during \`hub::draw\`, then read back in \`render::draw\` after the hub renders so the placement push step doesn't have to re-derive layout geometry.
- Stable per-slot UUIDs (\`0x4C40_7000_…\` for trophies, \`0x4C40_5470_…\` for the celebration) let \`TerminalImageRenderState::build_commands\` dedupe consecutive frames so we don't re-emit Sixel bytes every tick.

## Scope

| | |
|---|---|
| Files | 8 modified + 2 new (\`hub/trophy_sixel.rs\`, \`hub/shop_celebration_sixel.rs\`) |
| Diff | +693 / −8 |
| New deps | none |
| New DB migrations | none |

## Test plan

- [x] \`cargo test -p late-ssh --lib -- hub::trophy_sixel hub::shop_celebration_sixel hub::shop::state hub::leaderboard hub::state\` — 20 tests pass (5 new trophy_sixel tests covering cache, display size, sixel-only-for-sixel, pixel coverage, palette distinctness; 3 new celebration_sixel tests; 1 new \`ShopPurchaseCelebration\` expiry test; existing hub tests still green)
- [ ] Manual smoke (WezTerm/Kitty/foot): open Hub Leaderboard — see gold/silver/bronze trophies on the top three rows of chips and arcade panels, score panels unchanged
- [ ] Manual smoke (WezTerm/Kitty/foot): open Hub Shop, buy an unlocked item — see a brief sparkle burst centered over the item list; equip / unequip an owned item — no burst
- [ ] Manual smoke (macOS Terminal / generic xterm): both surfaces render exactly as today, no errors


mpiorowski:
In general, I’d like us to avoid rendering full-quality images in the terminal unless there’s a strong reason. This is a server-side SSH app, and pushing full image payloads through terminal escape sequences is expensive, fragile, and can make the UI feel bad. For small decorative effects, we should prefer text/ASCII, tiny fixed-size assets, or skip the effect entirely.

Also, I am planning on rewriting whole badge system anyway ;p

PR #279, opened 2026-05-29
feat: pixel sayonara scene in the quit-confirm modal

hardlygospel:
Thanks Mat — small additive sixel drop on the quit-confirm modal in the same lane as #277 and #278. One feature, one new file, no business logic touched.

---

## Summary

When the quit-confirm modal opens, capable terminals (Sixel / Kitty / iTerm2) get a small pixel sunset with a waving silhouette overlaid in the spacer area between the prompt and the footer. Non-capable terminals see only the existing \"Clicked by mistake, right?\" prompt and the q / Esc footer — exactly as today.

The aim is just \"when you're about to leave, the room waves back.\" Pure flavour, ~3 seconds of attention while you decide whether to actually quit. No new persistent state, no DB changes, no broadcast — placement is computed from the existing modal geometry and read by the renderer after \`quit_confirm::ui::draw\`.

## Implementation notes

- Scene is drawn procedurally with \`image::RgbaImage\` (no bundled PNG assets), PNG-encoded once per protocol, cached for the lifetime of the process.
- Routed through the shared \`terminal_image::terminal_image_from_bytes\` pipeline that already backs every other terminal-image surface — same wipe semantics (#254), same protocol gating, same fallback shape as the cup ritual from #277.
- Stable per-frame UUID lets \`TerminalImageRenderState::build_commands\` dedupe consecutive frames so we don't re-emit bytes every tick while the modal is open.
- The placement area is computed via a new \`quit_confirm::ui::sayonara_scene_area(modal_area)\` helper. It returns \`None\` when the modal can't fit the scene cleanly (e.g. very small terminals), so the non-image path handles edge cases automatically.

## Scope

| | |
|---|---|
| Files | 4 modified + 1 new (\`quit_confirm/sayonara_sixel.rs\`) |
| Diff | +302 / −1 |
| New deps | none |
| New DB migrations | none |

## Test plan

- [x] \`cargo test -p late-ssh --lib -- quit_confirm\` — 5 tests pass (4 new sayonara_sixel tests covering cache, display size, sixel-only-for-sixel, meaningful pixel coverage; existing quit_confirm input test still green)
- [ ] Manual smoke (WezTerm / Kitty / foot): press q twice to open quit confirm — see the small pixel sunset scene with a waving figure between the prompt and the q / Esc footer
- [ ] Manual smoke (macOS Terminal / generic xterm): same flow, see only the existing text prompt + footer, no errors


mpiorowski:
sorry for killing all the img prs :D

PR #280, opened 2026-05-29
feat: Add chess piece pngs

ricott1:
Replaces the hand-drawn ASCII chess piece art with png pieces.

Three render modes, cycled with `v`:
  - `png`: full-resolution PNG via Kitty/iTerm2/Sixel terminal-image protocols.
  - `8x8`: Half-block fallback driven by the 8×8 piece PNGs (similar to Asterion/sshattrick rendering).
  - `ascii`: the prior `_+_ /K\` silhouettes.

**NOTE**: `image_render` is moved in #276, so we would need to change the import accordingly when that's merged.

### `png`
chess_large

### `8x8`
chess_small


mpiorowski:
Thank you!

PR #285, opened 2026-05-30
bump russh to 0.61.1, drop pkcs5 pin

ricott1:
The pkcs5 = "=0.8.0-rc.13" workaround was a temporary pin for russh 0.60.1's prerelease pkcs8 stack. russh 0.61.1 resolves cleanly against pkcs5 0.8.0, so the pin (and the late-ssh pkcs5 workspace dep that anchored it) are removed. CONTEXT.md note also dropped.


mpiorowski:
goat, thank you

PR #300, opened 2026-06-02
feat: room-scoped slash commands

### LINKED ISSUE #291: FR: room-scoped slash commands
## Description

Right now all slash commands are global. The list in CHAT_COMMANDS is the same no matter which room you're in. Adding some infrastructure so that commands can be scoped to a specific room (or room type), so a room can expose its own set of commands on top of the global ones could allow to have room-specific features without clogging global commands.

### Is your feature request related to a problem?
As we add more themed/interactive rooms, it's getting awkward that every command shows up everywhere. A command like /sheet only makes sense inside the dnd room, but with the current setup we'd either have to make it global (and have it pollute the command list + do nothing in unrelated rooms) or not add it at all.

### What solution would you like?
- A room can declare a set of commands that are only available/auto-completed when you're in that room.
- The command composer / autocomplete merges global commands with the current room's commands.
- Commands typed in a room that doesn't own them are ignored.

### Do you have any additional context?
The current global command list lives in CHAT_COMMANDS (late-ssh/src/app/chat/state.rs), and the composer is started per-room already (start_slash_command_composer / start_command_composer_in_room), so the room context is available at the point where commands are offered. This seems like a natural place to hook in room-scoped commands.

ricott1:
- Closes #291.
- Add temp example with /sheet command in dnd room.


ricott1:
I refactored commands to a new file and moved them to a new struct with a CommandScope enum. I think this pattern could generalize nicely in the future, e.g. we can support commands for room type easily if needed

mpiorowski:
@ricott1 love this refractor, thank you

PR #325, opened 2026-06-07
Voice: status icons + live speaking indicator

hardlygospel:
## What

The voice roster used to show each participant as `@name <word>` with the status as plain text, and the controls line was text-only. This adds **status icons** and a **live speaking indicator**:

- Each participant row gets a status icon: 🟢 speaking · ⚪ listening · 🔇 muted · 🔕 deafened.
- A speaking user's **name pops green + bold** while they're talking — a real-time "who's talking" cue.
- The controls line gains matching icons: `🟢 live · ⏎ leave · 🎤 u mic · 🎧 d deafen`.

## How

Presence is factored into a small priority-ordered enum (`deafened > muted > speaking > listening`) with `icon()` / `label()` / `color()`, so the roster and the controls line stay in sync and the logic is unit-testable.

## Tests
Two new unit tests cover the presence precedence and that every state has a distinct icon and label. `cargo test -p late-ssh --lib voice` is green; no new clippy warnings.

UI-only change, contained to `voice/ui.rs`. No service, schema, or config changes.

## Thanks 🙏
- **mat (@mpiorowski)** — for late.sh and the voice feature this builds on. Thank you for the foundation that makes polish like this possible.

### DIFF
--- a/late-ssh/src/app/voice/ui.rs
+++ b/late-ssh/src/app/voice/ui.rs
@@ -104,16 +104,14 @@ pub fn draw_voice_controls(frame: &mut Frame, area: Rect, view: &VoiceRoomView<'
     } else if !view.paired_cli_supports_voice() {
         "Run the native late CLI to join voice.".to_string()
     } else if let Some(participant) = view.snapshot.participant(view.current_user_id) {
-        let state = if participant.deafened {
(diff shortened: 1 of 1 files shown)


hardlygospel:

PR #326, opened 2026-06-07
Voice: moderator kick/unblock from the mod console

hardlygospel:
## What

Voice had **no moderation path** — a disruptive user in the call couldn't be removed. This adds two commands to the `/mod` console:

- `kick voice @user [reason]` — removes the user from voice now and blocks them from rejoining.
- `unban voice @user [reason]` — lifts the block.

## How it's enforced

The shared `VoiceService` gains a runtime block set:

- `kick()` drops the user from the roster **and** records the block. Enforcement is the **LiveKit token gate**: `join_ticket` refuses a blocked user, so they genuinely can't publish or subscribe — not just hidden from the roster. Self-reported presence from a blocked client is also dropped.
- `allow()` lifts the block.
- The block is **runtime-only by design** (clears on restart). A DB-backed `voice_ban` table with durations could be a follow-up if you'd like persistence — I kept this PR migration-free.

## Wiring

- New caps `KICK_FROM_VOICE` / `UNBLOCK_VOICE`, granted to moderators, gated **per target tier** like every other surface, and written to the moderation **audit log**.
- Folded into the existing `kick` / `unban` grammar and `/mod` help text (`help kick voice`, `help unban voice`), so it reads exactly like the server/room/audio actions already do.
- The shared `VoiceService` is threaded into `ModerationInfra` alongside the existing artboard handles.

## Tests

- `VoiceService`: token-gate denial while blocked, roster removal, and that a blocked client re-reporting presence stays out.
- Command parser: `kick voice` / `unban voice` parse to the right action and require a target.

`cargo build -p late-ssh` and the voice + moderation test suites are green; no new clippy warnings.

## Thanks 🙏
- **mat (@mpiorowski)** — for late.sh and the voice + moderation systems this slots into. The mod console and permission tiers made this a clean, small addition. Thank you.

### DIFF
--- a/late-ssh/src/app/voice/svc.rs
+++ b/late-ssh/src/app/voice/svc.rs
@@ -6,7 +6,7 @@ use late_core::MutexRecover;
 use serde::Serialize;
 use sha2::Sha256;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt,
     sync::{Arc, Mutex},
 };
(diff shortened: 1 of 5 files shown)


hardlygospel:

PR #327, opened 2026-06-07
Voice: per-room voice flag — make a room VC or text-only

hardlygospel:
## What

Voice is currently a single server-wide channel with no per-room control. This adds the **control plane to make a room VC or text-only**:

- `room-voice #room on` / `room-voice #room off` in the `/mod` console.
- `view #room` now shows the room's `voice: on/off` flag.

## How

- **Schema:** new `chat_rooms.voice_enabled` column (migration `074`), **defaults to `false`** so every existing room stays text-only until a moderator opts it in. Model gains the field and a `ChatRoom::set_voice_enabled` setter.
- **Command:** folded into the moderation module exactly like `rename-room` — new `RoomVoice { slug, enabled }` command, a new `SET_ROOM_VOICE` capability granted to moderators, an audit-log entry, and help text (`help room-voice`).

## Scope / follow-up

This is the **per-room control + data layer** — the part that makes "a room is VC or not" a real, mod-managed property. Giving each enabled room its **own separate call** (room-keyed LiveKit rooms instead of the one shared channel) is a deliberate follow-up: voice presence and join/listen tickets are currently keyed to one fixed room over the native CLI websocket and the browser listener HTTP endpoint, so true per-room calls need matching changes on both clients. I kept this PR server-side and migration-light so the flag can land and be built on.

## Tests

- Command parser: `room-voice #x on/off` parse correctly and reject missing/invalid state.
- Existing chat/room-search suites updated for the new model field and stay green.

`cargo build` + the moderation/chat/room-search suites pass; no new clippy warnings.

## Thanks 🙏
- **mat (@mpiorowski)** — for late.sh, and for the voice + moderation foundations this extends. The room model and mod-command framework made this a small, clean addition. Thank you.

### DIFF
--- a/late-core/migrations/074_add_chat_room_voice_enabled.sql
+++ b/late-core/migrations/074_add_chat_room_voice_enabled.sql
@@ -1,0 +1,4 @@
+-- Whether a room is voice-enabled (VC) or text-only. Defaults to false so every
+-- existing room stays text-only until a moderator turns voice on for it.
+ALTER TABLE chat_rooms
+    ADD COLUMN voice_enabled BOOLEAN NOT NULL DEFAULT false;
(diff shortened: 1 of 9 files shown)


hardlygospel:

PR #341, opened 2026-06-12
[testers-wanted] Add embedded IRC server (ircd)

mevanlc:
- Introduces a lightweight pure-rust IRC server (ircd) that exposes late-sh's internal chat as an IRC-client-compatible surface
- The IRC server logic was written from scratch, custom for the late-sh repo
    - Reusing an irc daemon crate would have made the PR smaller by LOCs but not smaller by complexity (and not higher quality, IMO)
- Feature is off-by-default unless enabled
- Dev recipes (e.g. make start) enable plaintext IRC by default -- see Makefile and .env for IRC related vars
- TLS is supported for encrypted connections (recommend TLS-only [no plaintext] in prod
	- The TLS certificate is not auto-setup so that is some up-front config required to get TLS going
	- Some IRC clients do chain-of-trust / domain-validation (like browsers w/ https), but most let you easily bypass these checks for self-certs
- Connecting to the late-sh ircd requires a password -- all major modern irc clients support configuring a connection password
	- Users start with no IRC password and thus their account cannot connect to IRC
	- Users desiring an IRC password can create a randomly allocated strong password in the late-ssh TUI under Settings->Account tab. The user can delete their password (thus disabling IRC for their account). The user can also regenerate a new password. Currently a user can have only 1 password at a time.
- User's IRC nickname is locked to their late-ssh TUI username and cannot be changed via IRC
- Most user-management moderation features map to IRC. Most notably: room kick, room ban, server kick, server ban
- Private channels are limited to members only
- Public channels can be `/list`ed and joined


mpiorowski:
OHHH, thats interesting!

PR #343, opened 2026-06-14
Shop categories and items left-click selectable with mouse

joaovitorbf:
This pull request makes shop category sub-tabs and item rows left-click selectable. Scroll wheel moves item selection. Buying remains keyboard-only (Enter).

- state.rs: added category_rects + item_rects storage with interior mutability (Cell/RefCell), hit-test methods, select_item and select_category_by_index
- ui.rs: draw_categories and draw_item_list now compute and store per-element Rects for mouse hit-testing
- input.rs: new handle_shop_mouse converts SGR coords to 0-based and dispatches clicks/scroll
- hub/CONTEXT.md: updated to document mouse behavior


mpiorowski:
than kyou!

PR #345, opened 2026-06-14
feat: Add 'Rebels in the sky' as a proxied game

ricott1:
## Summary

Adds **Rebels in the Sky** as a late.sh door game. late.sh opens an outbound SSH
connection to the standalone rebels server (`frittura.org:3788`, configurable), runs the remote terminal stream through an embedded `vt100` emulator, and renders it as a ratatui widget in its own page (tab `7`).

We preserve the top bar so it looks like we're still in late.sh, because of this we need the vt100 emulator rather than just copy/pasting raw bytes into the buffer.

## How it works

- **Outbound proxy** (`app/door/rebels/proxy.rs`): a per-session `russh` client connects to the rebels server, requests a PTY + shell, and bridges bytes both ways.
- **Identity** (`identity.rs`): a stable per-account `(username, Ed25519 key)` is derived from the late.sh `user_id` with blake3 and used for publickey auth, so the same account always maps to the same rebels save.
- **Embedded rendering** (`render.rs`): remote output feeds a `vt100::Parser`, blitted into the widget each frame; the proxy reader wakes the session's `RenderSignal` on new output for prompt repaints.
- **Input**: while the game is running, all keys and mouse events forward raw to rebels. Exit by quitting rebels (**Esc** then confirm, or **Ctrl-C**); any disconnect returns to the Launcher.

## Configuration

`LATE_REBELS_ENABLED` (default on), `LATE_REBELS_HOST` (`frittura.org`),
`LATE_REBELS_PORT` (`3788`), `LATE_REBELS_SECRET` (seeds the derived identity).
When disabled, the page shows "Currently unavailable", another option would be to not display the page at all but I thought that it would get messy if we add more pages later.

## Dependencies & testing

- New deps: `vt100` and `blake3`.
- Unit tests: identity derivation, `vt100`-to-`Buffer` blitter, mouse-offset clipping. Integration (`tests/rebels_proxy.rs`): stub `russh` server covering connect -> PTY -> shell -> output round-trip and disconnect-to-Launcher.


mpiorowski:
GOAT

PR #350, opened 2026-06-17
Enhance Minesweeper with solid blocks, mouse support, and grid layout

joaovitorbf:
This pull request adds mouse support to the Minesweeper game in the arcade section and improves the Minesweeper UI for better clarity and interaction. It also fixes the issue with having multiple lives where you hit a mine but the mine count doesn't decrease, messing up the counting of mines when playing hard boards.

Left click to open a cell or chord. Right click to flag an unopened cell.

Old board:
image

New board:
image


mpiorowski:
I love the grid layout, but would drop the solid blocks, or make it optional, switchable via key.

Can we do it? @joaovitorbf :) and thx for more contributions!

joaovitorbf:
Oh yeah, definitely. I added screenshots exactly because I thought it was going to be controversial ahhaha. 

Do you think there should be an ingame toggle or maybe an option in the setting's tweaks tab? 

I'll also try just the grid with no blocks later today, haven't tried that, maybe it already is good enough for visibility.

mpiorowski:
in game toggle for sure, but make the current dot one a default one, and then people can swtich to full blocks

joaovitorbf:
I've added a cell style toggle with the "o" key, defaults to the dot style. I'd say the dots look pretty good with the grid!

Kooha-2026-06-17-16-43-52  <br>   
image

PR #387, opened 2026-06-29
chat: idle presence ranks, a new-account link guard, and /mod timeout

hardlygospel:
Thanks @mpiorowski! Two small community-health features for chat, both off `main`.

## ⏳ Idle presence ranks
Users now accumulate **connected time** (a new `users.online_seconds` column, migration `092`), accrued per session on disconnect (capped at 24h/session to shrug off clock weirdness). That earns an **idle "presence rank"** shown as a badge beside the name — you rank up just by hanging around:

| Rank | | Connected |
|---|---|---|
| Drifter | 🪶 | 1h |
| Regular | ☕ | 24h |
| Local | 🛋 | 100h |
| Fixture | 🏛 | 200h |
| Landmark | 🗿 | 1000h |

The rank **rides the existing chat-badge channel** (any purchased flag/badge is kept after it), so it renders everywhere names do — SSH and web — with no new plumbing.

## 🚫 New-account link guard
**Brand-new accounts (younger than 3 hours) can't post links** — a common spam-and-leave vector. It **escalates** rather than punishing at once:
- **1st link** → the author is **warned** and the message is **dropped** before it's ever stored or broadcast (the link is never shown, so nobody can click it). *No timeout yet.*
- **2nd link** → a **5-minute posting timeout**, on top of the warning (an in-memory mute that also blocks further attempts).

The detector catches `http(s)://`, `www.`, and bare `domain.tld` forms. **Admins and moderators are exempt.**

## 🔇 `/mod timeout` — quick fixed-interval timeouts
A new mod-room command for a fast server timeout without typing free-form durations:

```
/mod timeout @name <5m|15m|1h|24h|2d|1w> [reason...]
```

It's a **durationed server ban under the hood**, so it reuses the proven ban path — ends the user's sessions, writes a moderation audit entry, **auto-lifts** when the interval elapses, and `unban server @name` lifts it early. Only the parsing and the friendly fixed-interval menu are new; an unknown interval is rejected with guidance. Documented in `/mod help` and `/mod help timeout`.

## Tests
```
cargo test -p late-ssh --lib app::chat::svc   # link detector + rank thresholds
cargo clippy / cargo fmt --check              # clean
```
The migration adds the column with a `DEFAULT 0`, so every existing account starts at zero and nothing else changes.

A couple of choices I'd love your steer on:
- Ranks accrue on **disconnect** (one DB write/session) rather than a periodic heartbeat — simplest, but a current session's time only lands when it ends. Web sessions accrue via the SSH path today; happy to wire `web_tunnel` too if you want parity.
- "Brand new" = **3h** and timeout = **5 min** per the brief; both are single constants if you'd prefer different values.


mpiorowski:
Closing, implementing a littler simpler version that dont need to hit db too much, jsut for mroe perf :) great idea nevertheless

PR #404, opened 2026-07-07
Add ticket system with modal UI, priority handling, and daily limits

odolha:
Mods can enable ticket support, users can add tickets (max 5/user/day).
Lists kept per room (each room has their own list).
We could initially enable for #bugs and #suggestions but it essentially can be used for anything.
Very basic functionality atm. Can be extended later.


mpiorowski:
Creating something similar, but simpler! thx for giving me the idea :)

PR #420, opened 2026-07-13
Super-snake - multiplayer snake-like game with some interesting features like edge warping; ported from my original Turbo-Pascal source code written in the 90s

odolha:


mpiorowski:
Added it as seperate pr :) too much conflcits with new lobby! GOAT WORK, thank you!

PR #438, opened 2026-07-19
Lateania: side-panel scroll counts wrapped rows, not logical lines

hardlygospel:
Thanks @mpiorowski! Small, self-contained bugfix for the Lateania side panels — most visible in the crafting menu that just landed in #425.

## The bug
The context side panel word-wraps (`Wrap { trim: false }`), so one logical line can occupy several terminal rows — but the scroll math counted **logical lines** and assumed one row each. In the 28–34-wide panel, any list whose rows wrap then drifts off the bottom and the last entries become unreachable. The crafting menu is the worst case: its ingredient/gated-reason rows routinely wrap, so at a multi-station room the last recipes (the cooking tier) are stranded below the screen. It's latent for shorter lists (long inventory/title rows) too.

## The fix
Count **wrapped rows** instead of logical lines: measure each line's height at the panel width (`wrapped_rows`, matching ratatui's `WordWrapper`), keep the selected line inside the visible rows with the existing scroll margin, and clamp so a manual scroll can't run into blank space. **Behaviour is unchanged when nothing wraps** — one line is one row, so the offset math reduces to exactly what it was before.

## Tests
- `wrapped_rows` word-wrap cases (fit, two-word wrap, an over-long word broken across rows, a real ingredient row)
- a regression test that the last entry of a wrapping list stays reachable
- the existing "selection stays visible in a long list" test, updated to the new signature

```
cargo test -p late-ssh --lib app::door::lateania::ui   # green
cargo clippy -p late-ssh --lib / cargo fmt -- --check   # clean
```

Branched off current `main`. Happy to adjust naming/scope however suits.

### DIFF
--- a/late-ssh/src/app/door/lateania/ui.rs
+++ b/late-ssh/src/app/door/lateania/ui.rs
@@ -370,8 +370,9 @@ fn draw_side(
     };
     let off = scroll_offset(
         state.list_scroll(),
-        lines.len(),
+        &lines,
         selected,
+        area.width as usize,
         area.height as usize,
     );
     state.set_list_scroll(off);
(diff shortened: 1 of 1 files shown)


hardlygospel:
Folded into #439 together with the collapsible-category panels — the scroll fix is the same commit there. Closing this in favour of the combined PR; happy to reopen if you'd rather review the scroll fix on its own.

PR #453, opened 2026-07-23
provide some username privacy and fun at the same time

mevanlc:
The #​1 upvoted comment on HN:
>I would prefer the connection instructions made clear that the default SSH username would be used to set up my username on the platform and display that to other users without initially seeking my consent. On $machine[work], I don't get to pick that $USER is my full name. If this affects you, make sure you connect with `ssh [pseudonymous name]@late.sh`.

Okay, maybe agree -- maybe not -- but they kind of have a point about surprise and privacy. So why not assign users a fun random nickname until they decide what they want to change it to themselves?  This PR does that.

We combine 353 curated "modifier" words with 605 curated "noun" words to give a total pool of 353 ⨉ 605 = 213,565 unique nicknames to randomly pick from.  (If somehow all nicknames get exhausted, we start appending numbers to them.)  A random & available name is assigned to new users at account creation time. That's it.  That's the change.

Here are some examples:
- PocketFirefly
- PrismaticPuffball
- PaperKestrel
- PatientMeteor
- SnugTortoise
- LinenTransistor
- MoonlitMoth
- SeawardPlover
- DewyDragonfly
- PutteringPager

Take it or leave it.  I had fun putting it together :)


mpiorowski:
I LOVE IT :D

PR #462, opened 2026-07-25
Shared coding scratchpad 

minavtafreshi:
Two people can now pair up and share a live text buffer. Both of you run `/pair @other` in chat within 10 minutes of each other, and you land on a shared full-screen editor. Esc leaves and lets the other person know you left.

Pairing is a mutual handshake on purpose. A one-sided `/pair` only leaves the other person a banner: it never changes their screen, never takes their keyboard, and re-running it will not ping them twice. Nobody can push state onto a session that did not ask for it.

Kept it small on purpose: no DB table (it dies once both people leave, same as the clubhouse lobby or the active-users list), no OT/CRDT merge, only two people at a time, and no syntax highlighting or line numbers. Concurrent typing from both sides can clobber, which is the accepted v1 boundary and is written down in the module's known gaps.


mpiorowski:
That is soooooo cool, let me play with it before merge. THANK YOU

PR #466, opened 2026-07-25
Highlight mentions and replies with a background wash

jeromeetienne:
A message that mentions you already got a colored bar in the left margin, and a reply to you got nothing at all. Both are easy to miss in a busy room. Give each one a background wash across the full row width instead, in two different colors so a reply and a mention stay tellable apart.

The two colors are derived by blending the palette's mention color and author color most of the way toward the canvas background, following the pattern SUDOKU_SAME_NUM_BG() already uses. That keeps every one of the 108 palettes working without adding a per-palette field. Palettes whose colors are not plain RGB fall back to the flat highlight background.

Detection lives in ensure_chat_rows_cache and is stored as a vector parallel to the painted rows, so the paint pass in visible_chat_rows is a straight index lookup. Three details:

- A reply is checked before a mention, because the composer prepends a `> @author: ...` quote line to every reply, which would otherwise make every reply to you look like a plain mention.
- Human replies store only the target message id (reply_to_user_id is filled in for bot replies only), so the target's author is resolved from the messages currently loaded. A reply whose target has scrolled out of the loaded window falls back to the mention wash rather than losing its highlight, because the quote line still names you.
- Your own messages never wash, so quoting or replying to yourself does not light up the room.

The mention test itself was `body.contains("@alice")`, which missed `@Alice` and falsely matched `@alicebob`. It now uses extract_mentions, the same parser the mention notification already uses, which is case-insensitive, respects word boundaries, and ignores mentions inside code spans.

Two smaller adjustments so the wash does not fight what was already drawn: the left margin bar is recolored to the reply accent on reply rows, and the selected-message marker now keeps the background of the row it replaces instead of punching a hole in it.


jeromeetienne:
Here is a little visual 

Screenshot 2026-07-25 at 11 16 21

jeromeetienne:
From here 

Screenshot 2026-07-25 at 11 19 34

PR #467, opened 2026-07-25
Add auto-idle nickname dimming with a per-user privacy opt-out

jeromeetienne:
## Summary

- Nicknames dim automatically after a configurable period of inactivity, set per-account in Settings -> Tweaks -> Idle -> "Auto-idle after (sec, 0=never)".
- A companion "Dim my name when idle" toggle lets each user opt out of showing their idle status to others at all. The privacy decision is resolved by the idle user's own session before publishing to the shared presence set, so an opted-out user's idle status is never visible to anyone else -- not even as a hint.
- A sleeping-face badge next to the username in chat mirrors the existing `/brb` moon badge, since the two are independent (a user can be both AFK and auto-idle at once).
- The dimmed look is produced by darkening the author's identity color directly (RGB scaling toward black) rather than relying on the terminal's SGR "faint" attribute, since many terminals don't render it reliably, especially layered on top of bold (own/friend author colors).
- Applies to chat, the voice roster, and the clubhouse nameplate.

## Test plan

- [x] `cargo check` / `cargo clippy -- -D warnings` / `cargo fmt --check` clean across `late-core` and `late-ssh`.
- [x] New unit tests in `late-ssh/src/app/tick_test.rs` cover the inactivity-to-idle transition, the privacy opt-out (idle locally but never published when the toggle is off), `0 = never` disabling auto-idle entirely, and instant clear-on-input -- all passing against a real Postgres test DB.
- [x] Full existing `late-ssh` test suite (2000+ tests) passes with no regressions.
- [x] Manually verified against two live SSH sessions: idle transition, instant un-dim on keypress, the privacy toggle hiding idle status entirely, and the badge/dimming rendering correctly in chat.


jeromeetienne:
- auto idle by default at 10min (per poll)
- auto idle may be disabled (for privacy reason)

Screenshot 2026-07-25 at 14 17 28

---

normal or dimmed on the map

Screenshot 2026-07-25 at 14 03 18
Screenshot 2026-07-25 at 14 03 12

---
normal or dimmed in the chat

Screenshot 2026-07-25 at 14 15 51
Screenshot 2026-07-25 at 14 16 10

mpiorowski:
Ok, I am parking this one, a little too much LOC and the value is not that high, the brb command takes care of most of the inactivity, and maybe other people prefer not to give it away. Sorry for it.

PR #472, opened 2026-07-26
feat(home): Exit qq footer hint + mouse-driven radio stage

hardlygospel:
Two small home-screen quality-of-life changes.

## Footer: "Exit qq"
Adds `Exit qq` after `Guide ?` in the home footer. No behaviour change: `qq` already exits (first `q` opens the quit confirm, second `q` confirms), so this just surfaces the binding next to the others.

## Mouse-driven radio/music stage
The music stage was keyboard-only (v1..v5, v+x, m). It's now usable and configurable with the mouse. The stage records its dock area each frame; `app::input` maps clicks to actions by the stage's fixed row layout:

- **Click a source header** (radio / youtube / icecast) to switch the paired-browser source.
- **Click a station/stream row** to select it (v1..v5 for radio, v1/v2 for icecast).
- **Click the mute row** to toggle mute.
- **Scroll over the stage** to change the paired client's volume.

New `App::set_paired_playback_source` sets a source directly; the existing `toggle_paired_playback_source` now delegates to it. Threaded a single `Cell<Option<Rect>>` (the dock area) through `SidebarProps`/`MusicStageProps`, mirroring the existing pet-strip rect pattern, so the input handler can hit-test without new per-element plumbing.

## Tests
- New sidebar test pins the dock row offsets the click handler depends on (mute/radio/youtube/icecast headers, v1..v5 radio, v1/v2 icecast), so a future re-layout can't silently break the clicks.
- `clippy`/`fmt` clean on the changed code. Full lib test suite compiles; the only failures are pre-existing Postgres-backed `audio::svc_test` cases that need a local DB.

Note: terminal mouse events couldn't be driven in my environment, so the offsets are verified by the guard test and the deterministic layout rather than a live click, worth a quick manual check on your end.

Thanks @mpiorowski for late.sh.

### DIFF
--- a/late-ssh/src/app/common/sidebar.rs
+++ b/late-ssh/src/app/common/sidebar.rs
@@ -105,6 +105,10 @@ pub(crate) struct SidebarProps<'a> {
     pub active_friend_names: &'a [String],
     /// Free-running frame counter for the music stage's marquee rows.
     pub marquee_tick: usize,
+    /// Slot the music stage records its dock area into each frame, so the mouse
+    /// handler in `app::input` can hit-test clicks (source headers, station
+    /// rows, mute) and scroll (volume) against it.
+    pub music_dock_rect: Option<&'a std::cell::Cell<Option<Rect>>>,
 }
 
 pub(crate) fn draw_sidebar(frame: &mut Frame, area: Rect, props: &SidebarProps<'_>) {
@@ -227,6 +231,7 @@ fn draw_sidebar_new_shell(frame: &mut Frame, area: Rect, props: &SidebarProps<'_
                         icecast_source_count: props.icecast_source_count,
                         radio_source_count: props.radio_source_count,
                         marquee_tick: props.marquee_tick,
+                        music_dock_rect: props.music_dock_rect,
                     },
                 );
             }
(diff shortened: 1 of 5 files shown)


mpiorowski:
Thanks for this, and sorry for the split answer.

**Taking the footer.** `Exit qq` is a good catch, the binding worked but nothing showed it. I'll pull that line in on its own.

**Leaving the mouse stage.** The approach is fine and I checked it properly: the dock rows really are fixed, and recording the rect each frame is the right way to handle the panel moving around. Three things would need fixing first:

1. `last_music_dock_rect` is never cleared. `App::render` has a clear list at the top (`render.rs:293`) for exactly this, and the pet-strip rects you copied are all in it. Without it the rect goes stale: switch to the Artboard, hide the sidebar with `Ctrl+O`, or resize small enough that the music panel gets dropped, and clicks there still toggle mute and switch source. On the Artboard that area is drawing canvas.

2. The scroll block sits above the `Screen::Dashboard` guard and skips the modal check the click path does. With Settings or the Hub open, scrolling there changes volume instead of scrolling the modal.

3. The test pins line indices while the handler uses dock rows. They only match because `Paragraph` draws line i at `dock.y + i`, and nothing covers that. `handle_music_stage_click` takes a plain `&mut App`, so you can test it directly without mouse events: set the rect, call it at `dock.y + 2`, check the source changed.

Past that, it's ~145 lines to put the mouse on a panel that already works from the keyboard. Not worth the surface area for me right now. That's my priorities, not your work, which is careful. Sorry to send you back.

PR #473, opened 2026-07-26
Stabilize adaptive tick tests under scheduler load

mevanlc:
- Settle based on elapsed clean time instead of tick counts
- Synchronize chat wake assertions with message creation events


mpiorowski:
goat

PR #484, opened 2026-07-28
feat(lateania): world map - path-into-fog hints + land/terrain/gather/champion indicators

hardlygospel:
Builds on the overhead world map (#475) and the corridors/off-map-arrow follow-up (#480, now merged). Two things: a fix for markers that looked stranded, and the map indicators to answer "which land is this and what's here".

## Path-into-fog hints
In play, a boss you'd discovered could show its marker with no visible way to it. That's fog of war: the map only draws a corridor between two rooms you've *both* explored, so a landmark whose approach still lies in the fog had nothing connecting it and read as "there's a boss here but no path to him".

- An exit from a visited room into the unknown now draws a faint direction arrow in the adjacent cell. Direction only, into an empty cell, so it never spoils what waits there - a found room just shows which way its paths run instead of looking stranded.
- Same treatment for the handful of same-level links the hand-authored core scatters too far apart to draw as a straight corridor.

## More map indicators
The map coloured rooms by biome but only marked bosses and taming spots, so you couldn't tell which land you were in or what an area held.

- **Land labels** - each explored region names itself near the centroid of its rooms in view (short form, e.g. "Embergate", "The Overworld"), placed only on a clear run of empty cells so the name reads whole and never hides a room or marker. Undiscovered lands stay unnamed.
- **Terrain key** - a footer line decoding the biome glyphs, listing only the biomes actually on screen so it stays legible.
- **Gather markers** - harvestable resource rooms show a pick, indexed from the world's resource nodes.
- **Champion markers** - the single toughest non-boss foe in each land shows a diamond. The endgame is wall-to-wall max-level mobs, so a per-room "elite" marker would carpet the map (4800+ rooms); one apex per region stays rare (~11). The crosshair inspector also now names the gather trade worked at a room.

Marker precedence: boss > tame > champion > gather > terrain, so a room shows its most significant feature.

## Tests
24 worldmap unit tests (added a fog-hint regression and a marker-index regression that pins the champion marker to stay rare). fmt + clippy clean.

Thanks @mpiorowski for the door and for merging the map work.

### DIFF
--- a/late-ssh/src/app/door/lateania/ui.rs
+++ b/late-ssh/src/app/door/lateania/ui.rs
@@ -284,6 +284,7 @@ fn draw_world_map(frame: &mut Frame, area: Rect, state: &State, view: &PlayerVie
                     Tile::Empty => (" ".to_string(), Style::default()),
                     Tile::LinkH => ("\u{2500}".to_string(), link_style), // ─
                     Tile::LinkV => ("\u{2502}".to_string(), link_style), // │
+                    Tile::Hint(ch) => (ch.to_string(), Style::default().fg(theme::TEXT_FAINT())),
                     Tile::Room(id) if *id == player_room => ("@".to_string(), player_style),
                     Tile::Room(id) => match poi(*id) {
                         Some(p) if p.boss.is_some() => ("\u{2605}".to_string(), boss_style),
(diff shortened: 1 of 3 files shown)


hardlygospel:
Folding this into #485 so the map/field work ships as one PR (per the same review preference). The two commits here (fog-path hints + land/terrain/gather/champion indicators) are cherry-picked into #485 unchanged, where the live walk-around field view builds directly on them.

PR #485, opened 2026-07-28
feat(lateania): live walk-around field view (RPG mode) + combat QoL + map indicators

hardlygospel:
A single Lateania door update: turn ordinary play into a walk-around rendered world while keeping the plain MUD a keystroke away, and smooth out combat. All terminal-native ASCII, works over any SSH client, no new engine or gameplay change. (Supersedes and includes #484.)

## Live walk-around field view ("RPG mode")
On a wide terminal the Room view gains a centre column: a scrolling, biome-coloured top-down field kept centred on you, log on the left, room prose on the right. Movement already walks room to room, so the field scrolls and the fog lifts ahead as you go. `M` toggles it; off, Lateania is the classic text MUD. The preference is persisted with your character. `m` still opens the separate pan-around overview atlas.

- **Artistic terrain, all nine biomes.** Each explored cell paints its biome as a hand-strewn scatter (grass with wildflowers, dark forest with trees, rolling water, ashen embers, cobbled town, cracked badlands, island shallows...), keyed to a stable per-cell hash so it never shimmers as you walk. Biome edges blend (grass gives way to cobble near a capital).
- **Caverns are carved from rock.** Underground inverts the fill: rooms and corridors are open floor, the space between is solid rock (▒▓█), so a dungeon reads as passages hewn through stone, not an open plain.
- **What's around you, colour-coded.** Live foes (red †) and other adventurers (cyan ☺) come from a fog-safe nearby list in the snapshot; shops/stations/stables/portals (⌂), bosses (★), tameables (♥), and resource nodes (♦) come from static tables at no snapshot cost. A colour key sits under the field.

## Combat quality-of-life
- **Q quaffs** the best healing potion in one keystroke without leaving the view (smallest potion that covers the wound, else the largest; nothing at full health).
- **Clickable action bar**: Attack, the class's ability slots, Quaff, Flee. Left-click runs the same action its key does. The `DoorGame` trait gains a default no-op `handle_mouse`; Lateania overrides it and hit-tests against absolute chip rects recorded at draw time (the pattern the mentions HUD uses), so there's no fragile draw-area reconstruction.
- **Click a foe** in the roster to lock onto that exact foe; falls back to auto-target if it's already gone.

## Overhead map indicators (was #484)
- Faint direction arrows for exits into the fog, so a discovered boss never reads as stranded (direction only, no spoiler).
- Land-name labels, a terrain key, gather-node markers, and a per-region "champion" (toughest non-boss foe) marker on the pan-around map.

## Tests
New unit tests cover: the quaff picker, action-bar chip layout + click hit-testing, click-to-target (incl. safe-haven refusal), nearby foes/players lists (with fog not spoiling hidden foes), the map marker index, and the persisted RPG-mode round-trip. fmt + clippy clean; the whole Lateania suite passes (the one failing test in the crate is a pre-existing DB-backed integration test that needs TEST_DATABASE_URL).

Thanks @mpiorowski for the door.


hardlygospel:
Folding this into #494 so everything I have open is one review - the field-view commits are contained there unchanged.

PR #487, opened 2026-07-28
feat: interaction mode (keyboard / mouse / hybrid) with first-run prompt + mouse clickability

hardlygospel:
Lets each person choose how they drive late.sh: keyboard-only (the classic terminal/programmer feel), mouse-first (Discord-like, click everything), or hybrid (both). A one-time prompt asks on first entry, then it lives in Settings.

## Interaction mode
- `InteractionMode` (keyboard / mouse / hybrid) persisted per-user in `users.settings` (`interaction_mode` key), following the existing theme/audio pattern; threaded through `SessionConfig` into `App`.
- **Mouse gate:** in keyboard mode the mouse-tracking DECSET sequences are withheld at session start (`enter_alt_screen`), so the terminal keeps its own text selection and copy; bracketed paste stays on. Mouse is on for mouse/hybrid.
- **Live toggle:** changing the mode flips terminal mouse reporting immediately (pushed onto the existing `pending_terminal_commands` queue) and persists, no reconnect needed. The top-level mouse dispatch also drops mouse events in keyboard mode as a belt-and-suspenders.
- **Settings:** Tweaks gains an "Input" row that cycles keyboard / mouse / hybrid.

## First-run prompt
After a new user enters late.sh (no saved mode), a one-time overlay asks how they want to get around: Both (default), Mouse, or Keyboard. Pick with 1-3, arrows + Enter, or a click (the mouse is live at first run so the prompt itself is clickable). It applies live, persists, and never returns; editable afterwards in Settings. New lightweight `app/onboarding/` module mirroring the quit-confirm pattern.

## Mouse clickability
Filled gaps so the common surfaces are fully mouse-driven (most already were):
- **Scratchpad:** click to place the caret, wheel to page the buffer.
- **Room search:** click a result to select and jump to that room.
- **Room info form:** click the Topic or Rules field to focus it.
- **Character sheet modal:** click Name/Body to focus.

Left keyboard-first by design: the mod modal (a command-typing interface), and the house-table card games + Green Dragon door — both render through wrapping/scrolling game panels where a click doesn't map cleanly to a menu row, and both are fully keyboard-playable.

## Tests
Cover the settings round-trip + mouse gate, the DECSET withholding, the onboarding options, and the scratchpad click-to-cell mapping. fmt + clippy clean; the whole suite passes bar the pre-existing DB-backed integration tests that need TEST_DATABASE_URL.

Thanks @mpiorowski for late.sh.

### DIFF
--- a/late-core/src/models/user.rs
+++ b/late-core/src/models/user.rs
@@ -41,6 +41,52 @@ impl AudioSource {
     }
 }
 
+/// How a session is driven. Chosen on first entry (see the onboarding prompt),
+/// then editable in settings. The key behavioural lever is whether the terminal
+/// mouse reporting is turned on: off in `Keyboard` so native selection/copy keep
+/// working; on in `Mouse` and `Hybrid`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum InteractionMode {
+    /// Keyboard only, the classic terminal/programmer experience; mouse
+    /// reporting stays off so the terminal's own text selection works.
+    Keyboard,
+    /// Mouse-first, Discord-like: everything is clickable, mouse reporting on.
+    Mouse,
+    /// Both keyboard shortcuts and the mouse work. The safe default.
+    #[default]
+    Hybrid,
+}
+
+impl InteractionMode {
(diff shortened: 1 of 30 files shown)


hardlygospel:
Folding this into #494 so everything I have open is one review - the interaction-mode commits are contained there unchanged.

PR #496, opened 2026-07-30
Add Bonsai Decay Shield shop item

jeromeetienne:
## Summary

Adds the Bonsai Decay Shield: a shop consumable that protects a user's bonsai tree from decay for two weeks.

Screenshot 2026-07-30 at 22 10 46


jeromeetienne:
from here 
Screenshot 2026-07-30 at 22 13 46

mpiorowski:
GOAT

PR #526, opened 2026-08-14
feat: Allow dropping items to make room for loot in A Dark Room

christippett:
Added functionality to drop items from your pack to make room for new items in A Dark Room. The mechanics/logic should match the implementation from the original game.

- Added `Phase::DropFor` and `Row::Drop` / `Row::DropCancel` to the event state machine. If selected loot exceeds free pack capacity, the game prompts "not enough room. choose what to drop:" and calculates the required quantity of carried items to discard (needed / item_weight).
- Dropping pack items transfers them to the ground loot list, picks up the target item and returns to the loot/story view so any dropped items can be retrieved later (if capacity permits).
- Added tests to `event_test.rs`


christippett:
Tacking on an extra commit that allows falling back to using fists in combat if there are no other usable weapons (e.g. if you're out of ammo).

mpiorowski (approved):
goat

PR #529, opened 2026-08-14
Add BashQuest as a door game, with tamper-resistant graduation certificates

hardlygospel:
```
 ██████╗  █████╗ ███████╗██╗  ██╗ ██████╗ ██╗   ██╗███████╗███████╗████████╗
 ██╔══██╗██╔══██╗██╔════╝██║  ██║██╔═══██╗██║   ██║██╔════╝██╔════╝╚══██╔══╝
 ██████╔╝███████║███████╗███████║██║   ██║██║   ██║█████╗  ███████╗   ██║
 ██╔══██╗██╔══██║╚════██║██╔══██║██║▄▄ ██║██║   ██║██╔══╝  ╚════██║   ██║
 ██████╔╝██║  ██║███████║██║  ██║╚██████╔╝╚██████╔╝███████╗███████║   ██║
 ╚═════╝ ╚═╝  ╚═╝╚══════╝╚═╝  ╚═╝ ╚══▀▀╝  ╚═════╝ ╚══════╝╚══════╝   ╚═╝
```

> **An original learning game by Tasmania** — 90 live challenges, eighteen tiers, one mentor who never lets you skip the hard part.

BashQuest is mine end to end: story, mentor, all 90 challenges. It's not a quiz — every answer is checked against real shell output from a real `bash`. You start at `ls`/`pwd`/`cd` and finish having built LVM storage, shared it over Samba and NFS, run a network with VLANs and a firewall, connected to a SAN over iSCSI, recovered from a kernel panic, built a kernel from source, run real version control and your own containers, and yes — made your desktop look genuinely good.

**Why it's more than a CLI cheat sheet**
- 🎯 **You run the actual command.** Answers are checked live against real output, not multiple choice.
- 🔥 **Streaks, lives, hints, skips — all tracked.** Push for a flawless run or lean on hints; it all shows up on your final stats.
- 🧑‍🏫 **One narrator, the whole way through.** Tasmania (Tony's in-game alter ego) talks you through every tier, every correct answer, every "nice try."
- 🎓 **A real graduation, not a completion toast.** Finish all 90 levels and Tasmania hands you a Certificate of Completion — and as of this PR, played through late.sh, that certificate is independently verified and published, not just a file sitting on someone's disk (more on that below).

**Eighteen tiers, ninety levels**

| # | Tier | What's in it |
|---|---|---|
| 1 | 🗺 Beginner | navigation, file ops, permissions, users |
| 2 | ⚙ Intermediate | processes, text processing, networking basics, shell scripting |
| 3 | 🔗 Pipes & Patterns | piping, I/O redirection, regex, `grep`/`sed`/`awk` |
| 4 | 🔧 Power Tools | `xargs`, `find -exec` |
| 5 | 🛡️ Expert | cron, logs, packages, compression, bash internals, systemd |
| 6 | 💾 Storage & Filesystems | disks, partitions, `mkfs`, `fstab`, LVM (volumes, resize, snapshots) |
| 7 | 📁 File Editing & Sharing | vim, ACLs, Samba, NFS, `rsync` |
| 8 | 🌐 Networking | IP addressing, routing, DNS, firewalls, VLANs, troubleshooting |
| 9 | 🔌 Storage Networking & SAN | iSCSI, NFS/SMB at scale, multipath, Fibre Channel |
| 10 | 🥾 Boot Process & Kernel | GRUB2, alt boot managers, kernel panics/recovery, building a kernel |
| 11 | 🎬 Media Management | ffmpeg, media library organization, a home media server |
| 12 | 🎨 Desktop Ricing | X11/Wayland, window managers, dotfiles, theming |
| 13 | 🌿 Git & Version Control | init/status/config, staging, history, branching, remotes, undo |
| 14 | 🐳 Docker & Containers | images, running/exec/logs, building, volumes/networks, Compose, cleanup |
| 15 | 📦 Universal Packages | snap install/channels/revert, flatpak install/remotes |
| 16 | 🪟 Terminal Multiplexing | tmux sessions, windows, panes, detach/reattach |
| 17 | 🎧 TUI Toolbelt | ranger, cmus, mpd/mpc, btop, fzf |
| 18 | 🌟 Modern TUI Tools | Yazi, lazygit, lazydocker, zellij |

That's the game. What this PR actually adds is bringing it onto late.sh as a real door, plus making "I finished it" mean something verifiable. Details below.

## Summary

- Adds BashQuest (my own 90-level, eighteen-tier bash training game) as a proper late.sh door, following the PTY-proxy pattern (like dopewars/DCSS/NetHack): a new `late-bashquest` host crate spawns bashquest.sh on a PTY behind SSH, and a new `late-ssh/src/app/door/bashquest` client connects to it like every other door, using the existing arcade-handle identity system.
- Adds a graduation certificate pipeline that's actually verifiable, not just a local text file: `late-bashquest` independently confirms bashquest.sh wrote a certificate on its own server-side storage (the player never gets shell access to fake it), reports it over the already-authenticated SSH channel via a host-originated marker nothing in-game can spoof, and late-ssh records it once per real account. A new public, read-only `late-web` endpoint lets a companion repo ([hardlygospel/bashquest-graduates](https://github.com/hardlygospel/bashquest-graduates)) mirror the verified list into a public gallery.
- Full design and chain-of-custody documented in `late-ssh/src/app/door/bashquest/CONTEXT.md`.

One deliberate scope cut, documented in both `terraform.yml` and the door's CONTEXT.md: `bashquest_image_tag` is optional rather than `required: true` like every other door, to avoid a coordinated change across all nine existing `deploy_*.yml` callers. Happy to follow up with that alignment separately if useful.

Thanks for maintaining late.sh, mat — this was a fun one to build on top of.

## Test plan

- [x] `cargo check -p late-bashquest -p late-ssh -p late-core -p late-web --all-targets` clean
- [x] `cargo test -p late-bashquest -p late-ssh -p late-web --lib bashquest` — all pass
- [x] `docker compose config` + Dockerfile stage cross-checks clean (no Docker daemon available to actually build)
- [x] YAML-validated every new/touched GitHub Actions workflow
- [ ] Manual launch/play/logout against a real deployed host (not possible from here — needs a live environment)
- [ ] `late-core/src/models/bashquest_graduate_test.rs` needs `TEST_DATABASE_URL` (a real Postgres) to run; not runnable in this sandbox, but mirrors the existing `game_payout_test.rs`/`arcade_handle_test.rs` harness shape


mevanlc:
#### some issues

0. defend against ^z wedging / suspending the game

1. save files are executable code

bashquest.sh uses: `source "$f"`.  Normal usernames are only stripped of whitespace, then used directly in the filename. A username such as `../payload` selects a path outside `.bashquest`, with `.save` appended. Writes also follow symlinks.  another process or user can place or modify the selected save file, logging in executes arbitrary shell code as the BashQuest service user.

2. Using regex meta characters in username can bypass the password check

Usernames are interpolated directly into grep regular expressions. For example, the username `.*$\|^x` turns login verification into a pattern whose first alternative matches any record, regardless of the submitted password.  This does not automatically load the matched other account's save file: the literal malicious username becomes PLAYER_NAME. But, it does make the password check succeed for an arbitrary identity. Also, colons are permitted in usernames, making users.db structurally ambiguous.

3. terminal escape injection through normal usernames

Normal usernames are allowed to contain backslashes and control characters. They are repeatedly interpolated into `%b` output. A username containing text such as `\033]`... is interpreted by printf `%b` as terminal control sequences. Depending on terminal support, this can alter the title, forge screen content, create deceptive hyperlinks, or manipulate the clipboard through OSC 52. Rejecting usernames that don't match 
```
\A[A-Za-z0-9_][A-Za-z0-9_.-]*\z
```
would make a lot of these `#2`/`#3` issues go away. Usernames with a . in them that get injected into a grep pattern would still need the dot backslash-escaped. Or switch to `grep -F`.

4. persistence does not defend against concurrency races

The script truncates the live save file and writes directly into it. Registration performs a check followed by a separate append. Concurrent sessions (multiple simultaneous users, or multiple ssh sessions under the same user -> door) can cause:

- duplicate account rows;
- overwriting of newer progress with stale state;
- load a partially written save;
- follow a pre-existing save-file symlink.

Most notably though, using a shared user db across all game sessions invites duplicates or interleaved data from concurrent/racing writes.

Even without the central user store, account creation (to, e.g., prevent 2+ users from creating accounts with the same name) is still vulnerable to TOCTOU races. A acquiring a shared lock file (with retry) could address this. The lock should not be held during password collection, however, as that invites a substantial shared resource bottleneck. Post-username collection and then again post-password collection would be correct and more friendly with lock-sharing.

Leaderboard reads could read half-written data. Writing to tmp and doing an atomic mv only after a successful write could ensure only fully-written files are visible to LB reads, but does not prevent stale sessions from overwriting newer progress. That requires a same-account session lease, CAS/versioning, or shared authoritative state.

5. unbounded input 

Prompt reads accept an unlimited line length. A user can send a very large username, password, or challenge answer and consume substantial memory in a Bash process.

6. Stack overflow possible in long sessions

Menus recurse instead of looping. A sufficiently long session or automated invalid-input stream continually grows the Bash call stack. At best this is wasteful on memory, in a worse case it might eventually lead to a stack overflow / OOM / core dump.

#### best practices 
1. password storage is unsalted md5

The script stores unsalted MD5 hashes. A copied users.db can be attacked offline extremely cheaply. The script also does not set a restrictive umask or explicit permissions, so .bashquest, hashes, saves, and certificates inherit the caller's defaults. Use a cryptographically secure salted hash and consider a restrictive umask.

2. doesn't need to be implemented in bash

Since the game never allows the user to exec "real" bash commands that are handled by the underlying OS (a Good Thing™), the game doesn't actually need to be written in the bash language. The game engine's bash-like pattern-matching-based simulation loop could be written in rust (or python, or ...) which is easier to secure, do robust error handling & recovery in, etc.

More interestingly than a mere language port of the existing logic: you could give a very authentic bash shell experience with an entirely rust-based bash port that virtualizes  absolutely everything, with basically zero chance of shell escapes, path traversal, or unsynchronized access to shared state (e.g. container fs / mounts).

- https://github.com/shantanugoel/rust-bash
- or for a more genuine but a more difficult adaptation, since it's usually used as a "real" shell replacement, not virtualization or sandboxing: https://github.com/reubeno/brush (see https://github.com/reubeno/brush/issues/1183)

---
3. The container hosting execution of the game shell script could have outbound network access disabled as a security backstop.  Shell command escapes or ACE in the container would therefore be unable to interact with other containers on the private network, or access the Internet.

4. Not an issue, but thought I'd point it out: since the user getting direct access to their save file in-game (or elsewise) would be a security model failure, the tamper-discouraging save-file obfuscation efforts are not really a feature late-sh needs (but no harm I guess).

hardlygospel:
Thanks for this, @mevanlc — went through it point by point and pushed fixes: [hardlygospel/bashquest@f8d269d](https://github.com/hardlygospel/bashquest/commit/f8d269d745a79e55153b15a08a3a737b61dbde60). This PR is now updated (commit above this comment) to pin that fixed commit with a re-verified checksum, so the door build picks it up rather than the old script.

**0. ^Z wedging** — `trap '' TSTP` added. Verified the process stays running (not `T`/stopped) after `kill -TSTP`.

**1. save files as executable code** — `source "$f"` is gone. `load_progress_safe` now reads `KEY=VALUE` lines against a fixed allowlist of keys, requiring non-negative-integer values; anything else on a line (including `$(...)`/backtick payloads) is silently skipped, never executed. Verified with a save file containing several command-substitution payloads — confirmed inert.

**2/3. regex-injection / escape-injection usernames** — usernames must now match `^[A-Za-z0-9_][A-Za-z0-9_.-]*$` (24 chars, checked before any lookup), and every lookup switched to `grep -F`/`awk -F:` field matching regardless. `:` is excluded from the charset too, so `users.db` can't be ambiguous. Verified `.*$\|^x` and a `\033]`-style payload both get rejected outright at the prompt now.

**4. concurrency races** —
- Save/users.db writes go through temp-file + atomic `mv`, which is both the atomic-visibility fix you described for the leaderboard *and* the symlink-write fix (`mv` replaces whatever the destination entry currently is rather than opening and writing through it). `load_progress` also now refuses outright to read through a pre-existing symlink.
- Registration TOCTOU: added a portable `mkdir`-based lock (no `flock` dependency, works on macOS+Linux) with a short-lived username claim, held only around the check-and-reserve step and again around the final append — never across the password prompt, per your suggestion. Verified under real concurrency: 5 simultaneous registrations for the same username → exactly 1 account created, 4 correctly told "taken."
- Left as-is, flagged rather than fixed: same-account concurrent-session overwrites. Agreed that's a session-lease/CAS problem, not a locking tweak, and out of scope for this pass.

**5. unbounded input** — every interactive `read` (username, password, menu choices, challenge answers) is now bounded.

**6. recursive menus** — the four menus that called themselves on invalid input (`main_menu`, `level_select`, `level_select_tier`, `startup_screen`) are now loops. Verified with 3000 consecutive invalid `main_menu` choices in one session — clean exit, no growth-related failure. I scoped this to just those four self-recursive spots rather than a full trampoline rewrite of every screen transition (the slower long-session mutual-recursion growth you're also right about is a real, bigger refactor I didn't take on here — noted in the commit message as a known residual).

**Best practices 1 (unsalted md5)** — new accounts get `openssl passwd -6` (salted SHA-512 crypt); existing unsalted rows still verify and are transparently upgraded to the new format on their next successful login. Added `umask 077` plus explicit `chmod` on the save dir, `users.db`, and every save/cert file.

**Best practices 2 (rewrite off bash)** — noted, not undertaken here. Appreciate the links; that's a real conversation worth having separately given the size of the lift.

**Best practices 3 (outbound network disabled in the container)** — that's an infra/deployment call outside `bashquest.sh` itself, flagging for whoever owns the door's runtime config rather than acting on it here.

**Best practices 4** — agreed, noted.

Full verification notes are in the [fix commit message](https://github.com/hardlygospel/bashquest/commit/f8d269d745a79e55153b15a08a3a737b61dbde60).

mpiorowski:
hey, got this running locally and played a good chunk of it. its fun, and the mevanlc fixes all check out in the pinned commit, i went through them one by one.

one thing i want fixed before it goes in: backspace doesnt work.

you type `ls x`, hit backspace, and it prints `^?` and the DEL char actually goes INTO your answer. arrows do the same thing, they stuff `^[[A` in there. so if you typo a command you cant fix it, you just submit garbage and eat the wrong answer. in a game about typing shell commands thats rough :D

its not late.sh doing it. i checked in a plain pty with nothing of ours in the way, typing `ls x` backspace backspace `-l`:

```
read -r -n 32 v     ->  v=[ls x\x7f\x7f-l]
read -e -r -n 32 v  ->  v=[ls-l]
```

its the `-n`. it takes bash out of canonical mode so it can stop after N chars, and then bashs own read loop doesnt do erase at all. `-e` hands the line to readline and it works. you keep the `-n` cap, which you want anyway since thats the unbounded input fix.

one gotcha if you do it: readline has to own the prompt. if you leave `printf "prompt"; read -e ...` then readline thinks its line starts at column 0, and anything that reaches for line start (ctrl-a, home, a redraw) sends a bare `\r` and starts overwriting your prompt. i hit exactly that. so the prompt needs to move into `read -e -p "..."`, with the color codes wrapped in `\001` `\002` so readline counts the visible width right. thats the 8 `read -r -n` spots. the 3 password ones (`read -rs -n`) should stay exactly as they are.

i did try patching this on our side with a sed and it made things worse, so we're back to shipping your script unmodified. would rather it live upstream anyway so the `curl | bash` players get it too, they have the same problem right now.

blocking on this one, everything else looks good.

mpiorowski (changes requested):
commented up

mpiorowski:
two more from reading the certificate pipeline. one blocks, one is just a "while youre in there".

**1. a failed db write loses the graduation forever. this one blocks.**

in `host.rs` the `.certificate.reported` sentinel gets written when the marker SEND succeeds:

```rust
if handle.data(channel, marker).await.is_ok() {
    let _ = std::fs::write(&reported, b"");
}
```

on our side `take_graduation()` removes the record, and `spawn_record` is fire and forget, it only logs if the insert fails. so if postgres hiccups for a second: no retry that session, and the sentinel means the marker never gets sent again in any later session either.

so someone finishes all 90 levels, we get one ERROR line in a log nobody reads, and they never show up in the gallery. no way to even notice it happened.

your own comment in host.rs says the db write is idempotent on account id anyway, so i think the sentinel can just go, or only get written after late-ssh says the row landed. then it fixes itself next time they log in.

**2. the marker isnt signed. not blocking, but cheap to do.**

the marker is `\0BQCERT\1 <blake3> \1 <handle> \1 <cert> \0`, and the blake3 is a hash of the certificate thats sitting right there in the same marker. so it proves the thing didnt get cut in half. it doesnt prove who sent it. theres no secret in it anywhere.

so "nothing in-game can spoof it" is really "nothing can print those bytes into the output stream". i checked, and today that holds: `user_input` never gets printed back, its only case matched and passed to chk/exact, and `PLAYER_NAME` does go through `printf '%b'` but USERNAME_RE keeps the weird stuff out. so its fine as it stands.

its just one line away from not being fine. `%b` turns `\0` and `\001` into real bytes, so the day a "you answered: X" line shows up with %b, a player can type their own marker and it gets published to a public gallery as a verified certificate.

both ends already share LATE_BASHQUEST_SECRET (thats where the ssh key comes from), so hmac the handle+cert with it and check it on our side. then it stops mattering what ends up in the stream.

rest of it reads well btw. the insert is race safe, the ON DELETE SET NULL reasoning is written down, every trust boundary has a comment saying why its trusted. its just this one ack, and a claim thats slightly ahead of the code.

hardlygospel:
Thanks for narrowing this all the way down, mat - the plain-pty repro made it a five-minute fix instead of a guessing game.

You had it exactly right: `read -n` drops bash out of canonical mode, so bash's own read loop does zero erase-key handling and backspace/arrows land as raw bytes in the answer. Switched every non-password `read -r -n` to `read -e -r -n`, moved the prompt into `read -e -p` so readline owns it (and hit your exact gotcha along the way - had to wrap the ANSI colour codes in `\001`/`\002` so readline's width math doesn't get thrown off by them and start overwriting the prompt on a redraw). The 3 password prompts (`read -rs -n`) are untouched.

Verified against a real pty this time (not just reading the diff): backspace, left/right arrows, and Ctrl+A + Delete all edit correctly with no prompt corruption, and the `-n` cap still auto-returns at the character limit exactly as before.

Pushed upstream: [hardlygospel/bashquest@61ed114](https://github.com/hardlygospel/bashquest/commit/61ed114360c0f749859598f4f06632bc011328de) - curl\|bash players get it too. Bumped the pin in this PR to match (commit above).

Also saw your certificate-pipeline comment (the failed-db-write / graduation-lost one) - haven't touched that yet, will pick it up next.

PR #538, opened 2026-08-17
fix: slash command detection

nopeless:
image

slash command detection should only trigger if slash is the first character in message


mpiorowski (approved):
goat

PR #540, opened 2026-08-17
Add savable daily chat logs, toggleable in settings

hardlygospel:
## Summary

- New "Save daily chat logs" toggle in Settings > Tweaks: when on, a plain-text transcript of the day's room chat and DMs is rebuilt at session end and saved to the R2 bucket already used for chat image uploads.
- New "View today's log" row opens an in-app read-only viewer (reuses the existing `Overlay` component chat already uses for member/reaction lists).
- The setting follows the same `settings JSONB` pattern as `start_with_music_muted` - no migration.

## Why the HMAC key

The R2 bucket is CDN-public by design (no ACL, unguessable UUID paths) - fine for chat images, not fine to reuse unmodified for private chat/DM content. So the log's object key is `logs/{hex(HMAC-SHA256(secret, "{user_id}|{date}"))}.txt`: deterministic (re-saving the same user+day cleanly overwrites) but infeasible to guess without a new server-held secret (`LATE_CHAT_LOG_SECRET`). The log is never surfaced to the client as a raw URL - it's always fetched and rendered server-side.

"Today" is the user's local calendar day (their profile timezone, UTC fallback), matching the "dated logs" framing.

## Infra

Added `LATE_CHAT_LOG_SECRET` end to end: `infra/secrets.tf` (new `random_password`/`kubernetes_secret_v1` pair, consumed only by service-ssh, not shared with another pod like the door-game identity seeds) and `infra/service-ssh.tf` (new env block), plus `.env.dev`/`.env.dev2` for local dev. **I can't run `terraform apply` or `terraform validate` from my sandbox (no terraform binary), so these two `.tf` files need a human look before merge.**

## Test plan

- [x] `cargo check -p late-core -p late-ssh --lib --tests` clean, no warnings
- [x] `cargo test -p late-ssh --lib settings_modal` - new `TweakRow::ALL` structural test + `chat_log_action_span` test pass
- [x] New `ChatMessage::list_for_daily_log` test added (`chat_message_test.rs`) covering room+DM inclusion, game-room exclusion, membership-gated exclusion, and date-range bounds - compiles clean but **could not actually run** in my sandbox (no Docker daemon available to bring up the test Postgres via `docker-compose.check.yml`)
- [ ] Manual smoke test (toggle on, chat, disconnect, reconnect, view log) - not possible without a live DB + R2 credentials in my sandbox, needs a human pass
- [ ] `terraform validate` on the two touched `.tf` files - needs a human with terraform installed

Thanks for taking a look, mat!


mpiorowski:
Love the idea and concept, but i have sth else in mind, will push a pr soon :) thank you though so much!

PR #550, opened 2026-08-21
Add persisted Le Word replay mode

mojoro:
## Summary

- add switchable daily and replay boards to Le Word, with a double-confirmed new-word action
- persist one replay board per user independently from daily progress
- keep replay answers distinct from the current and daily answers, without daily wins, chips, or quest credit

## Changes

- migrate Le Word progress to separate `daily` and `replay` slots while retaining the legacy uniqueness constraint for rolling deploys
- restore both modes during session bootstrap and serialize queued saves so rapid board transitions persist in order
- update Le Word controls, help copy, workspace behavior, and architecture documentation
- cover mode persistence, reconnects, answer rotation, reset confirmation, save ordering, and reward isolation

## Test Plan

- [x] `cargo fmt -p late-core -p late-ssh -- --check`
- [x] `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 cargo clippy -j 8 -p late-core -p late-ssh --all-targets --features otel -- -D warnings`
- [x] `make test-llm ARGS="-p late-core -p late-ssh -E 'test(le_word)'"` (19 passed)
- [ ] `make check` — blocked on this host before project code compiled because `late-webview` requires the unavailable system package `gdk-3.0.pc`

## Notes for Reviewers

The migration intentionally retains the legacy `(user_id, puzzle_date)` uniqueness constraint so old pods can finish in-flight daily saves during a rolling deployment; replay rows use a nullable puzzle date.


mevanlc:
It's tempting to reuse the `le_word_games` table for this because it's got mostly the right shape. However, making a non-additive change to `le_word_games` would require a phased/deferred rollout model to correctly handle the fact that late.sh can have old pod(s) running alongside new pod(s).  Therefore I suggest a new table, `le_word_replay_games`, which is an additive-only change, that is largely `le_word_games`-shaped (but not entirely identical, which further justifies its existence.)

```
  CREATE TABLE le_word_replay_games (
      id UUID PRIMARY KEY DEFAULT uuidv7(),
      created TIMESTAMPTZ NOT NULL DEFAULT current_timestamp,
      updated TIMESTAMPTZ NOT NULL DEFAULT current_timestamp,
      user_id UUID NOT NULL UNIQUE
          REFERENCES users(id) ON DELETE CASCADE,
      answer_word TEXT NOT NULL CHECK (answer_word ~ '^[a-z]{5}$'),
      guesses JSONB NOT NULL DEFAULT '[]'::jsonb,
      current_guess TEXT NOT NULL DEFAULT '',
      is_game_over BOOLEAN NOT NULL DEFAULT false,
      won BOOLEAN NOT NULL DEFAULT false
  );
  ```
  
  
 The Rust side could have distinct models: `DailyGame` (just a rename of 'Game'), and `ReplayGame`. Also:
 
```
struct BoardSnapshot {
      answer: String,
      guesses: Vec<String>,
      current_guess: String,
      is_game_over: bool,
      won: bool,
  }

  struct State {
      daily_snapshot: Option<(NaiveDate, BoardSnapshot)>,
      replay_snapshot: Option<BoardSnapshot>,
      // ...
  }
```

The cost is duplicated table columns and two persistence methods (which is cheap, IMO.) The benefit is migration simplicity and modeling "similar but not identical" things separately instead of as a Chimera.

mpiorowski:
I am gonna park this pr, not sure i want a repeatable mode for worlde, you could theroricallty try out all our words combinations, "train" them ;p

PR #552, opened 2026-08-22
A suggested modification to PR#550

mevanlc:
This PR builds on [mpiorowski/late-sh#550](https://github.com/mpiorowski/late-sh/pull/550). It targets that PR's feature branch rather than `main`.

  ## Summary

  - Preserve the existing `le_word_games` schema for daily progress keyed by `(user_id, puzzle_date)`.
  - Add `le_word_replay_games` for one replaceable replay board per user.
  - Separate daily and replay models, save commands, snapshots, and session preloads.
  - Load daily progress using the exact date returned with the selected daily word.
  - Cancel pending replay-reset confirmation when leaving Le Word or opening global help.
  - Restore the visible `?` help hint.

  ## Rationale

  Replay state has different identity and retention semantics from daily state. Keeping it in a separate table avoids rewriting the daily table, weakening its typed `puzzle_date`, or replacing its existing uniqueness constraint.

  Because the new table is additive, the replay feature can be deployed without a destructive migration or a two-phase application/database rollout.

  ## Verification

  - `make check`
  - `cargo_lint`
  - `cargo check -p late-core -p late-ssh`
  - `cargo clippy -p late-core -p late-ssh --all-targets --no-deps -- -D warnings`
  - `cargo nextest run -p late-core -p late-ssh le_word --no-fail-fast` (21 passed)
  - `git diff --check`


mevanlc:
Sorry folks -- still learning how to do stacked PRs:

Closing in favor of mojoro/late-sh#1, which is the correctly stacked PR targeting #550’s `feat/le-word-replay-mode` branch.

Both PRs have the same head commit (`a44d47c`), so no proposed changes are being discarded; #1 simply presents the intended delta against #550 rather than against `main`.

PR #569, opened 2026-08-29
feat: Improve responsiveness in Super Snake

christippett:
When playing Super Snake I noticed that the inputs felt sluggish and laggy, especially when compared to the single-player version.

### The Root Cause
Because Super Snake is a multiplayer game governed by a shared server-side tick loop (running at 125ms intervals), it could not safely use the single-player "accel" trick (which instantly moves a snake out-of-band on a keypress) without breaking collision fairness between players. 

However there are two areas in the current implementation that contribute to the feeling of input lag:
1. **Direction Overwriting:** When a player steered, their input simply overwrote the `motion` state. If a player executed a rapid double-turn (e.g., turning `Up` then `Left` to navigate a tight corner within a single 125ms window), the `Up` input was instantly overwritten by `Left`, causing the snake to miss the corner entirely.
2. **Timer Drift:** The server loop used a naïve `sleep(125ms)` followed by a mutex lock. Under contention, this caused the tick rate to jitter and stretch, leading to micro-stutters.

#### 1. Per-Player Input Queues
Introduced a bounded input queue (`VecDeque<Direction>`, capped at 2) to each `PlayerState`. 
* Instead of overwriting the snake's direction, valid turns are now buffered.
* The game loop pops exactly one queued direction per tick.
* **Impact:** Players can now rapidly roll their fingers across the arrow keys (e.g., a quick U-turn), and the server will faithfully execute the sequence across consecutive ticks. No inputs are dropped.

#### 2. Advanced Reversal Guards
The anti-reversal logic (preventing a snake from reversing into its own neck) was updated to validate incoming keypresses against the *deepest queued direction* rather than the last executed move. This guarantees players can safely queue complex maneuvers without accidentally triggering self-collision guards.

#### 3. Strict Interval Cadence
Replaced the drift-prone `sleep` loop with `tokio::time::interval_at` configured with `MissedTickBehavior::Skip`. 
* **Impact:** The game now ticks at a rigid, unwavering cadence. If the server experiences a brief lag spike, Tokio absorbs the delay and realigns to the fixed schedule rather than compounding the drift or bursting catch-up frames. This makes the snake's movement visually smooth and predictable.


mpiorowski:
THANK YOU :)

PR #600, opened 2026-09-13
Add a careers page (placeholder, no open role yet)

hardlygospel:
## Summary

Thanks for late.sh, mat — this one's speculative, feel free to close it if it's not something you want.

- New static Askama page at `/careers` (+ `/careers.xml` feed), following the `legal/terms.html` pattern exactly.
- No real opening exists right now, so the one posting shipped is clearly labelled "example — not an open role" on the page.
- Postings carry a `live` bool. Only `live` postings get `schema.org/JobPosting` JSON-LD embedded (what Google's and LinkedIn's free job-search crawlers actually key off — no API/partner access needed) or appear in the XML feed. This is deliberate: emitting structured job data for a placeholder would mislead those crawlers and risks the domain's structured data getting flagged as spam. Flip a posting to `live: true` with real details whenever/if there is one.
- Added a `/careers` card to the homepage next to `/profiles`.
- Updated the routes table in `late-web/CONTEXT.md`.

## Test plan

- [x] `cargo check -p late-web` passes, and `cargo test -p late-web` (all 35 tests, including the 3 new `careers_test.rs` cases) is green
- [ ] Visual check in a browser — couldn't spin up the local Docker stack in this environment, worth a look before merging


mpiorowski:
I am not sure we are at the size where this is needed. From the moment Ive launch this app, only ONE person asked me if there is a carrier page.

mpiorowski:
Ok, taking your ideas from here and putting them on the late app :) as alwasy thansk for goat work

PR #601, opened 2026-09-13
Mouse drag to reorder sidebar panels

hardlygospel:
## Summary

Thanks mat.

- The sidebar panel editor (Settings > Tweaks > Sidebar panels) could only reorder rows with `[`/`]`. This adds mouse drag as a second way to do the same thing: Down on a row grabs it, dragging over a different row's rect moves it there immediately, Up releases.
- Reuses `move_right_sidebar_component` as-is (same Vec-splice + autosave keyboard reorder already calls) — drag and `[`/`]` share one code path, so there's no risk of them drifting apart.
- Row rects are cached each frame the same way the settings modal already caches its tab-strip rects (`tab_rects`/`tab_at_point`), just applied to the panel-editor row list.
- Scope note: I looked at whether this could extend to freely dragging panels around the live home-screen sidebar (not just inside the editor), but that list is order-based, not position-based, and there's no floating-panel concept anywhere in the app to drag into — reordering inside the existing editor was the real, already-shipped surface for this. Happy to hear if you had something more ambitious in mind here.

## Test plan

- [x] `cargo check -p late-ssh` (incl. `--tests`) passes
- [x] Added `rect_list_index_at_point` unit tests (row hit-testing, misses, hidden rows) — pure-function tests, not yet run through the workspace test suite in this environment
- [ ] Manual drag test in a real terminal client — I don't have a way to drive real SGR mouse-drag sequences against a running server in this environment, worth trying before merge


mpiorowski:
Closing this. The drag itself works, but the cost is out of proportion to what it buys.

With love!

PR #607, opened 2026-09-15
Improvements to gallery/artboard web view

christippett:
Add pinch-to-zoom, pan and tap-to-fit touch gestures to make the gallery more accessible for mobile users.

Also added the Noto Sans font that should cover off some of the missing unicode symbols that are currently not showing.


mevanlc:
Awesome, thanks for doing this!

mpiorowski:
goat

